PR #5, opened 2015-02-13
Update index.php

fly-in-red:
missing the last php ending.


matecat:
Sorry, we rejected your pull request:

http://php.net/manual/en/language.basic-syntax.phptags.php

###### If a file is pure PHP code, it is preferable to omit the PHP closing tag at the end of the file. This prevents accidental whitespace or new lines being added after the PHP closing tag, which may cause unwanted effects because PHP will start output buffering when there is no intention from the programmer to send any output at that point in the script.

PR #14, opened 2015-08-19
Engine error reporting during translation.

### LINKED ISSUE #10: Tilde MT integration
I'm opening a new issue to use for communication on Tilde MT integration.

Here is a preview of basic Tilde MT integration [tilde-nlp/MateCat/commits/lets-mt-integration](https://github.com/tilde-nlp/MateCat/commits/lets-mt-integration). This basically just follows the stuff that was already there for MicrosoftHub and Moses engines.

The above code works but the user has to enter three different ID's (Client ID, System ID, Terms ID) to add our engine. So we would like to display some basic config options after the user enters the Client ID so that the System ID and Terms ID can be selected from a list. Here is a preview on how this might look [tilde-nlp/MateCat/commits/letsmt-advanced-integration](https://github.com/tilde-nlp/MateCat/commits/letsmt-advanced-integration).

I add the possibility to specify another template to be displayed after the user presses _Confirm_ in the initial view. The template is then filled with data that are returned by the engine (in place of the id of the newly added engine that would be returned in the usual scenario).

The stuff that is not yet implemented is some custom JavaScript that would be attached to the System selection dropdown onchange event (in the newly added engine config view). This is necessary to be able to return the available term collection ID's after the user selects a system from the dropdown. When the user clicks an element in the dropdown an AJAX call would be made to fetch the relevant term list and fill out the term dropdown.

This is the basic idea. Does this seem reasonable?

### LINKED ISSUE #12: Displaying engine error messages during translation
Error messages that are returned by an MT engine during translation aren't shown to the user.

This is a problem in the case of the Tilde MT engine. If a Tilde translation system has not been used for a while it enters a stand-by mode. It gets brought up again after a new translation request is made but waking up a system might take up to 10 minutes. This is a very common scenario so there really should be a way to let the user know why he isn't getting any translations from his MT engine. He might not be getting a translation due to a system still waking up or a system could have been turned off or deleted without the user knowing (by some other user). So there should be a way to let the user know whether he should just continue working while the system gets brought up or should he not expect a translation without some further actions (e.g. going to the Tilde MT site to check on the system).

Are there any plans to implement a mechanism to show error messages or some king of notifications during translation?

rihardsk:
Closes #12, closes #10
This adds on top of #13 some error reporting functionality in the document translation view.

Firstly we make [the rest of the response](https://github.com/matecat/MateCat/commit/4a739a5f39d34256e6a63f26b8110ee60c09b7be#diff-d9e66f5a209f37f869b77d6faede18dfR144) available when an error is returned to a request. This allows us to [pass engine errors to the frontend](https://github.com/matecat/MateCat/commit/76f4c11f36700cc85bc95829c4b039787377cbba#diff-350226ecb327b1389944f5046f36c3b1R229) where [they are rendered](https://github.com/matecat/MateCat/commit/76f4c11f36700cc85bc95829c4b039787377cbba#diff-0016513769a46f73f3b1104e87eb4237R6378) bellow the list of translation suggestions if the error message has either the code `-1001` or `-1002`.

### DIFF
--- a/lib/Controller/getContributionController.php
+++ b/lib/Controller/getContributionController.php
@@ -225,6 +225,8 @@ class getContributionController extends ajaxController {
             $mt_result = $mt->get( $config );
 
             if ( isset( $mt_result['error']['code'] ) ) {
+                $mt_result['error']['created_by_type'] = 'MT';
+                $this->result[ 'errors' ][] = $mt_result['error'];
                 $mt_result = false;
             }
 
@@ -333,7 +335,6 @@ class getContributionController extends ajaxController {
         }
 
         $this->result[ 'data' ][ 'matches' ] = $matches;
-
     }
 
     private function setSuggestionReport( $matches ) {
--- a/lib/Utils/Engines/AbstractEngine.php
+++ b/lib/Utils/Engines/AbstractEngine.php
@@ -135,16 +135,14 @@ abstract class Engines_AbstractEngine {
         if ( $mh->hasError( $resourceHash ) ) {
             $curl_error = $mh->getError( $resourceHash );
             Log::doLog( 'Curl Error: (http status ' . $curl_error[ 'http_code' ] .') '. $curl_error[ 'errno' ] . " - " . $curl_error[ 'error' ] . " " . var_export( parse_url( $url ), true ) );
+            $responseRawValue = $mh->getSingleContent( $resourceHash );
             $rawValue = array(
                     'error' => array(
                             'code'    => -$curl_error[ 'errno' ],
-                            'message' => " {$curl_error[ 'error' ]}. Server Not Available (http status " . $curl_error[ 'http_code' ] .")"
+                            'message' => " {$curl_error[ 'error' ]}. Server Not Available (http status " . $curl_error[ 'http_code' ] .")",
+                            'response' => $responseRawValue // Some useful info might still be contained in the response body
                     )
             ); //return negative number
-            $responseRawValue = $mh->getSingleContent( $resourceHash ); // Although an error was returned some useful info might still be contained in the response body
-            if(!empty($responseRawValue)) {
-                $rawValue['result'] = $responseRawValue;
-            }
         } else {
             $rawValue = $mh->getSingleContent( $resourceHash );
         }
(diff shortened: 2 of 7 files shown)


Ostico:
Hi @rihardsk,

sorry but i've some issues to expose to you:

* [this file](https://github.com/matecat/MateCat/commit/76f4c11f36700cc85bc95829c4b039787377cbba#diff-0016513769a46f73f3b1104e87eb4237R6378) is no more present in MateCat repository, it is builded in production environment with grunt task runner, so all your changes must be done on the files in [this directory](https://github.com/matecat/MateCat/tree/master/public/js/cat_source). Please update to the last MateCat version and refer to [this guide](http://www.matecat.com/advanced-manual-setup/#cjc) on how to compile your javascript code in order to test it.

* i seen an critical issue when a customer use minimum_qe value:
[Check for QE value](https://github.com/tilde-nlp/MateCat/blob/error-messages/lib/Utils/Engines/LetsMT.php#L57). The engine class should raise an error code [Here](https://github.com/tilde-nlp/MateCat/blob/error-messages/lib/Utils/Engines/LetsMT.php#L64) instead of nullify the translation value, because that behaviour create troubles to the [analysis daemon](https://github.com/matecat/MateCat/blob/master/lib/Utils/Analysis/tmAnalysisThreadChild.php).

* [This minus sign](https://github.com/tilde-nlp/MateCat/blob/error-messages/lib/Utils/Engines/LetsMT.php#L165) will convert all your error messages to a "0" ( zero )

* Please use different codes rather than 1001 and 1002 ( ex: 2001 and 2002 ), because the 1000 family code is used or could be used internally by MateCat:
[-1000](https://github.com/matecat/MateCat/blob/master/inc/Bootstrap.php#L183)

* Other issues are in the comments to your code. A different solution should be implemented.

rihardsk:
Fixed the mentioned issues.

> this file is no more present in MateCat repository, it is builded in production environment with grunt task runner, so all your changes must be done on the files in this directory. Please update to the last MateCat version and refer to this guide on how to compile your javascript code in order to test it.

In 621024f794645746a63d51c3f65a3a2b0691c854

>i seen an critical issue when a customer use minimum_qe value:
Check for QE value. The engine class should raise an error code like this instead of nullify the translation value, because that behaviour create troubles to the analysis daemon.

In f74ab337e91062d0d16598f742fcb5359dd9add6

>This minus sign will convert all your error messages to a "0" ( zero )

In 0a9b69c31bdb5f0406ee6fbb9358a5c15bc7a908

>Please use different codes rather than 1001 and 1002 ( ex: 2001 and 2002 ), because the 1000 family code is used or could be used internally by MateCat:
-1000

In 6e0c94fa88c7f2efc85f4a8c9739ca017e61f5d4

a2cacfc67bc13aa7372b2cd8cc4468efc99e7b0d adds document language checking to translate requests. This is according to the suggestion in the Google doc that Domenico sent.

PR #51, opened 2016-04-14
Second stage of Matecat - LexiQA UI integration. Live QA checking

dfoteinakis:
Please update the lexiqa.matecat.com instance.

Thank you,
Dimitris


Ostico:
Please can you merge master branch into this branch?

Thank you

dfoteinakis:
Ok, I will as soon as i verify that what I just submitted works and send
you a new pull request.

Thank you!

Ostico:
Hi @dfoteinakis 
i seen you added a javascript library for the tooltips, we already use:

 - jquery.tooltipster.min.js

Can you substitute your library with this? Tnx
Update it if needed.

dfoteinakis:
We are aware that tooltipster is already in matecat source and we tried to
use it but it did not work for us at the time. Changing our code to using
it, is not straightforward and will take valuable time.

We would like to have the updated UI up and running as soon as possible in
order to get some feedback on the actual UI and QA performance. Thus, if us
using powertip for now, is a blocking issue for you, please let us know. In
that case we will have to setup our own instance and delay the merge.

Thank you!

Ostico:
Obviously this is not blocking strictly speaking, but what would happen if we allowed all to add libraries without any control?
How many different libraries we can have to do the same thing?

In the past we suffered some of slowdown in the cattool dued to some libraries introduced. 
Many users/translators does not have powerful pc.

I hope you will work taking in consideration this fact.

dfoteinakis:
Yes of course I understand, and unless we hit a major deadend, we'll try to
change to tooltip stet. what I submitted is by no means set in stone. In
fact I have removed the console.log stripping from grunt because we might
have other issues we have not seen yet and may have to debug from the
browser.
Since the integrated UI is still under development and review is pending,
we'll probably need to make changes once our and your people play with it.
That's why I would prefer not to use up time now to change the library,
especially since we still adding new features.

PR #58, opened 2016-05-02
Fix tests to run on php 5.3

freegenie:
@dinies @Ostico this fixes tests in php 5.3, maybe the name of variables I used is meaningless. Advice welcome.


dinies:
Nice, I agree with your choice for the name.

PR #72, opened 2016-05-09
Fixed getSegments query to have MySQL 5.6/5.7 compatibility

Ostico:
Fixed getSegments query to have MySQL 5.6/5.7 compatibility


freegenie:
Awesome thanks!

PR #74, opened 2016-05-10
Bug job analysis

riccio82:


Ostico:
@freegenie can you check that this fix do not breaks anything on your side?

PR #85, opened 2016-05-12
LexiQA v.1.0

dfoteinakis:
This merge is up to date with matecat/master branch (11th May).
The LexiQA module can be enabled/disabled by changing the config parameter we added: lxq_enabled.
This is currently set to true (in index.html)
We have tested this on our own instance at http://matecat.lexiqa.net


composer.json
```diff
@@ -4,7 +4,7 @@
     },
     "require": {
         "php": ">=5.3.0",
-        "robmorgan/phinx": "^0.4.6",
```
Ostico: Phinx update changed symphony dependencies to php >= 5.5

https://github.com/matecat/MateCat/pull/85/files#diff-1da2c7edc898c70e5a79a9997c98ceccR361

This upgrade broken our server configuration because of our different version of PHP
There is a reason why you updated it?

dfoteinakis: No, it updated automatically, when running the upgrade scripts.
I probably should have excluded it from the commit, but it slipped through.

Do you want me to issue a fresh pull request?

Ostico: No tnx, since there is not a specific reason i just reverted the file .

Ostico:
There is a bug on the LexiQA bar in the translate page. when clicking on the files list, the dropdwon slides under the bar:
image

If you fix that issue, please submit your pull request against LexiQA branch please not master. I will close this pull request, i already merged in LexiQA branch so, start from such branch.

dfoteinakis:
Ok will do.

Ostico:
There is another issue, but it is critical. Seems that the SetTranslation is no more sent to the server.
image

This is our master:
image

Ostico:
Nevermind, seem related to an internal error for a wrong configuration.

PR #136, opened 2016-05-31
Pee table bugfix

Lyonard:
.. and added a draft function pee2payable


Ostico:
Please squash unuseful and repeated commit points...

PR #147, opened 2016-06-03
fix highlight of segment after render

freegenie:
Refactor of the previous pull request that was breaking search. This is a more lightweight refactoring. 
Highlight happens inside UI.scrollSegment, after scrolling.


freegenie:
closing to republish as hotfix

PR #185, opened 2016-06-14
New panel integration

freegenie:
Server side API endpoint to save tag_projection, speech2text, lexiqa preference per chunk. 
Project wide preference can be overridden at chunk level.


freegenie:
@Ostico I'm closing this to rebase on develop. There is also some test that needs to be conducted yet to verify conflicts between lexiqa and matcat's glossary and blacklist checks.

PR #191, opened 2016-06-14
Hotfix propagation

FrancescoE:


freegenie:
I wonder if sprintf is as safe as PDO params. This solution doesn't look correct to me.

Ostico:
Sure it is not the best solution but works for now. 
The right solution is to create ? placeholders as many id are in the array and call bindValue on the PDOStatement in a foreach statement.
It must be rewritten. Anyway, the list of IDs is get from a previous query, so it is safe enough.

freegenie:
Got it. Anyway, I guess the bug is mine. Sorry guys!

freegenie:
I turns out that I'm not able to reproduce this bug.

Ostico:
To reproduce, you must have a document with more than one ID to propagate.
The PDOStatement transform the id imploded string in int and only the first id is evaluated.

Anyway, see this gist:
https://gist.github.com/Ostico/16afa402733ff3b6d15df9edf3ddad78

freegenie:
Thanks now I got it. The problem with the tests I did is that it _seem_ to work properly in UI, because UI takes a different path from the server side process. The UI is updated in place, and also the word count show on page looks correct. This is only visible when you reload the page.

freegenie:
BTW now that we have a list of propagated IDs, I think it would be better for the setTranslation to return the propagated IDs and have the UI update only those segments. This is likely to reduce and simplify client side code too.

Ostico:
I'm working on it now

PR #201, opened 2016-06-17
Fix rebutted in revise_improved

edgareler:


public/js/cat_source/es6/react/ReviewTranslationIssueCommentsContainer.js
```diff
@@ -6,36 +6,87 @@ export default React.createClass({
...
+            undoRebutVisible : false
         }
     },
 
     replyClick : function() {
         this.setState({ replying: true });
     },
 
+    undoRebutClick : function() {
+        ReviewImproved.setIssueRebutted( this.props.sid, this.props.issueId, 'false' );
```
freegenie: why need to pass this false as string @edgareler ?

freegenie: moreover, better to pass an object with key value map. `false` as third argument is not very meaningful in function named `setIssueRebuttted`

public/js/cat_source/es6/react/ReviewTranslationIssueCommentsContainer.js
```diff
@@ -44,10 +95,28 @@ export default React.createClass({
...
+        this.setState({ rebutLabel : 'Sending', rebutDisabled : true, sendDisabled : true });
+
+        var data = {
+          message : this.state.comment_text,
+          source_page : (config.isReview ? 2 : 1)  // TODO: move this to UI property
+        };
+
+        this.setState({ rebutLabel : 'Sending', rebutDisabled : true });
+        ReviewImproved.submitComment( this.props.sid, this.props.issueId, data );
+        ReviewImproved.setIssueRebutted( this.props.sid, this.props.issueId, 'true' );
```
freegenie: Also here, why pass 'true' as string? 
multiple ajax requests that represent a unique transaction should be avoided when possible. When not avoidable, put them in a `$.when` function so we can chain `done` and `fail` functions that trigger only when both succeed or when at least one fails.

PR #207, opened 2016-06-20
Disable lexiqa when qachecks active

edgareler:


lib/Controller/newProjectController.php
```diff
@@ -369,4 +371,29 @@ private function generateGDriveAuthUrl(){
         $this->gdriveAuthUrl = \GDrive::generateGDriveAuthUrl();
     }
 
+    private function isToDenyLexiQA() {
+        $database = \Database::obtain();
+
+        $userDao = new \Users_UserDao( $database );
+        $user = $userDao->getByUid( $_SESSION[ 'uid' ] );
```
freegenie: @edgareler I wonder if you can use $this->logged_user which is already set by a parent constructor

lib/Controller/newProjectController.php
```diff
@@ -369,4 +371,29 @@ private function generateGDriveAuthUrl(){
...
+
+        if( $user != null ) {
+            $ownerFeatureDao = new OwnerFeatures_OwnerFeatureDao($database);
+
+            $isQaGlossaryEnabled = $ownerFeatureDao->isFeatureEnabled(
```
freegenie: @edgareler about this, these are two separate queries, maybe we can implement something clever to do the same using `FeatureSet` class. Something like

```
$featureSet = FeatureSet::fromIdCustomer( $id_customer ); 
$fearureSet->isEnabled( $code ); 
```

edgareler: Yes, this will remove the demand for two queries.

PR #211, opened 2016-06-20
Hotfix for review improved

freegenie:
A set of fixes for review improved to handle a case when submit was dying with no error message and button was stuck with _Sending_ label.


Ostico:
@freegenie tell me if this code must be produced ASAP.

Ostico:
Promoted

PR #221, opened 2016-06-21
Fix scroll to first segment of each project's file

FrancescoE:


freegenie:
@FrancescoE I'm pretty sure this is breaking a desired behaviour. Let's talk about it.

Ostico:
@FrancescoE , i asked you if this was be tested, but if @freegenie has some doubt about your patch, talk to him and let me know what's your conclusion.

PR #262, opened 2016-07-04
Fix speech2text

freegenie:


public/js/cat_source/speech2text.js
```diff
@@ -74,19 +73,15 @@ Speech2Text.init  = function () {
...
-
-                    var segmentId = segment.data('split-original-id');
-                    var segmentRecord = MateCat.db.segments.by('sid', segmentId);
 
-                    Speech2Text.isToEmptyTargetElement = Speech2Text
-                        .shouldEmptyTargetElement(segmentRecord);
+                    Speech2Text.targetElement = Speech2Text.microphone.parent().find('.editarea');
 
-                    Speech2Text.microphone.click(Speech2Text.clickMicrophone);
+                    Speech2Text.microphone.on('click', Speech2Text.clickMicrophone );
```
freegenie: We prefer to use 'on' to set the event explicitly instead of the click shortcut, more clear to read.

public/js/cat_source/speech2text.js
```diff
@@ -100,29 +95,36 @@ Speech2Text.init  = function () {
...
+                var segmentId = new UI.Segment( segment ).absId ;
+                var segmentRecord = MateCat.db.segments.by('sid', segmentId);
+
                 if (!microphone.hasClass('micSpeechActive')) {
                     microphone.addClass('micSpeechActive');
                     Speech2Text.animateSpeechActive();
                 }
 
-                if (Speech2Text.isToEmptyTargetElement) {
+                if (Speech2Text.shouldEmptyTargetElement( segmentRecord )) {
```
freegenie: Instead of relying on a global value, I pass the segmentRecord around, because this is triggered by a specific event and passing the specific object as argument is always better when possible, instead of relying on global state.

public/js/cat_source/ui.contribution.js
```diff
@@ -39,15 +39,26 @@ $.extend(UI, {
...
-			$(editarea).html(translation).addClass('fromSuggestion');
+            // XXX we are modifing the APP state so that MateCat will know that the object is changed
+            // in particular this is needed for the Speech2Text to know that the newly added text is coming
+            // from a 100% match.
+            var segmentObj = MateCat.db.segments.by('sid', UI.getSegmentId( segment ) );
+            segmentObj.suggestion_match = match.replace('%', '');
+            MateCat.db.segments.update( segmentObj );
+
+			$(editarea).html( translation );
+            $(document).trigger('contribution:copied', { translation: translation, segment: segment });
```
freegenie: Added this event in UI contribution. This is necessary to understand when contribution got copied, to notify Speech2text and keep code decoupled. See like I'm passing the copied object, alway preferable instead of reading it back inside the event callback.

public/js/cat_source/ui.contribution.js
```diff
@@ -39,15 +39,26 @@ $.extend(UI, {
...
 			this.saveInUndoStack('copysuggestion');
 			$('.percentuage', segment).text(match).removeClass('per-orange per-green per-blue per-yellow').addClass(percentageClass).addClass('visible');
             $('.repetition', segment).hide();
 			if (which) {
 				this.currentSegment.addClass('modified');
-                                this.currentSegment.data('modified', true);
-                                this.currentSegment.trigger('modified');
-                        }
+                this.currentSegment.data('modified', true);
+                this.currentSegment.trigger('modified');
```
freegenie: This is sick, class, data and trigger, we'll need to refactor sooner or later.

public/js/cat_source/ui.core.js
```diff
@@ -2564,13 +2567,43 @@ UI = {
                 }
                 UI.execSetTranslationTail();
 				UI.setTranslation_success(data, this[1]);
+
+                var record = MateCat.db.segments.by('sid', data.translation.sid);
+                MateCat.db.segments.update( _.extend(record, data.translation) );
```
freegenie: This is very important.  Speech2text relies on the state of segments collection, we change the segment state here, this should be the preferred way to modify app state in the future. Better than using css classes, jquery data, or DOM attributes.

public/js/cat_source/ui.core.js
```diff
@@ -2564,13 +2567,43 @@ UI = {
...
+     * This function is an attempt to centralize all distributed logic used to mark
+     * the segment as modified. When a segment is modified we set the class and we set
+     * data. And we trigger an event.
+     *
+     * Preferred way would be to use MateCat.db.segments to save this data, and have the
+     * UI redraw after this change. This would help transition to component based architecture.
+     *
+     * @param el
+     */
+    setSegmentModified : function( el, isModified ) {
```
freegenie: Self explanatory, read the comment.

PR #268, opened 2016-07-05
Change mark as completed

freegenie:
Introduces a lock ( readonly ) for translate page when revision is started, and a warning on revision side when translation is running.


freegenie:
Closed to reapply as HotFix.

PR #283, opened 2016-07-08
Do not take into account db update for splitted segment

freegenie:
This resolves a problem that happens on splitted segments, even if dictation is not enabled.


public/js/cat_source/ui.core.js
```diff
@@ -2295,8 +2295,8 @@ UI = {
 	checkVersion: function() {
```
Ostico: TODO:
We should implement a out-of-band communication for the version number ( returned to the client ) and trigger this function to notify the users that the version is changed.

This function works ( is called ) only on page reload and only when the javascript is not reloaded at the right version by the browser.

PR #294, opened 2016-07-12
Fix splitted document error

riccio82:


freegenie:
bug was introduced here cf1aa60fcdde63ac7e8af41cb26a74fbe18b851e

PR #300, opened 2016-07-13
Rework getLocalizedName function for lang validaiton

freegenie:
This fixes a case that was observed in development environment: for unknown reasons some the cookies we use to store the source and target langs for history happen to be full lowercase ( e.g. it-it). This breaks the page due to the validation we recently introduced. 

This also refactors a little the `doAction` method in controller.

### DIFF
--- a/inc/Bootstrap.php
+++ b/inc/Bootstrap.php
@@ -350,8 +350,9 @@ class Bootstrap {
      */
     public static function areMandatoryKeysPresent()  {
         $merged_config = array_merge(self::$CONFIG, self::$CONFIG[ INIT::$ENV ] );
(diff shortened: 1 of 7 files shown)


freegenie:
@Ostico there may be broken development environments until this is merged. Me and @riccio82 had problems with manage page.

Ostico:
The language cookie is not used in the manage page if i remember

freegenie:
indeed, but I had to do other changes for the manage page. The function that raises exception was renamed, and the previous that was more permissive is used whenever we have to get input from database, where RFC version may not always be used. ( see projects created via API  ).

PR #326, opened 2016-07-29
Encrypt oauth token

edgareler:


freegenie:
nice PR, I wonder if we really need another config file and not just use a mandatory key in config.ini? Like wordpress does, prepopulate it in the sample with a default string to change and instructions to make it more secure.

edgareler:
@freegenie we can put the key in the config.ini (or maybe oauth_config.ini). This PR is creating a new key file at [DefuseEncryption->createEncryptionKey()](https://github.com/matecat/MateCat/blob/develop/lib/Utils/DefuseEncryption.php#L49-L70) when no key is found. This process can be done in the config.ini file either, it's just a matter of choice the more suitable solution.

Issue #309, opened 2016-07-18
The @ upload modifier should be replaced by CURLFile object above php 5.6.


yifanjiang:
It was aware that the curl option CURLOPT_SAFE_UPLOAD is set to true by default:

```
lib/Utils/Filters.php-41-            if ( PHP_MINOR_VERSION >= 5 ) {
lib/Utils/Filters.php-42-                /**
lib/Utils/Filters.php-43-                 * Added in PHP 5.5.0 with FALSE as the default value.
lib/Utils/Filters.php-44-                 * PHP 5.6.0 changes the default value to TRUE.
lib/Utils/Filters.php-45-                 */
lib/Utils/Filters.php:46:                $options[ CURLOPT_SAFE_UPLOAD ] = false;
lib/Utils/Filters.php-47-            }
```

However it doesn't really help much for a php5.6 environment. e.g. trying to upload a file from the upload page (front page) will lead to a file uploaded with unexpected file content, in which only the file path prefixed by an '@' instead the real file content resides.

So I guess something like this will be more usable:

```
- lib/Utils/Filters.php:            'documentContent' => "@$filePath",
+ lib/Utils/Filters.php:           if ( PHP_MINOR_VERSION >= 6 ) {
+ lib/Utils/Filters.php:                'documentContent' => new \CURLFile($filePath),
...

```

Ostico:
@yifanjiang tnx, we will update with this patch

SimZal:
@yifanjiang Can you confirm that the fix works for you?

yifanjiang:
@SimZal Thanks for the fixing :-) It looks great! I could hardly manage a environment to do it last week, but I'll try to verify the fix very soon!

yifanjiang:
Verified by cherry-pick to 1.0.5. Thank you!

PR #349, opened 2016-08-21
Fixed version check for PHP > 5

SimZal:
1) Version check
Existing code is relying only the constant PHP_MINOR_VERSION (i.e. 2 for PHP 5.2). This works for PHP 5, but when major version comes out (PHP 7.0) the code thinks it's running on 5.0 because it's not checking for major version.

I've added the check for major version to fix the problem. 

2) Your VirtualBox image includes PHP 5.6 and uploading to mashape doesn't work (issue #309 ) with it, because the use of CURLFile is becoming mandatory (because of security).

### DIFF
--- a/lib/View/popup_google_login.html
+++ b/lib/View/popup_google_login.html
@@ -6,7 +6,7 @@
 
         <h1>Login with Google</h1>
 
-        <p>Log in to access advanced functionalities<br />(e.g. TM, MT, glossary management <br /> and Google Drive integration)</p><br/>
+        <p>Log in to access advanced functionalities<br />(e.g. TM, MT, glossary management <br /> and Google Drive integration)</p>
         <div class="popup-box">
             <div class="outer">
                 <div class="inner">


lib/Utils/Engines/MyMemory.php
```diff
@@ -319,7 +319,7 @@ public function glossaryImport( $file, $key, $name = false ) {
 
         $postFields[ 'key' ] = trim( $key );
 
-        if ( PHP_MINOR_VERSION >= 5 ) {
+        if ( PHP_MAJOR_VERSION >= 5 && PHP_MINOR_VERSION >= 5 ) {
```
Ostico: Hi @SimZal 

i loked into your code and this patch will works fine only for PHP versions:

> 5.5 <= PHP < 6.0
> 6.5 <= PHP < 7.0
> PHP >= 7.5

but not for the rest of versions.

PR #364, opened 2016-08-26
CurlFile support for PHP >= 5.5

SimZal:
I added a new function to Utils to be used any time Matecat wants to upload a file with Curl.
I think this commit fixes the issue #309. I'm unsure if there are other places where Matecat uploads files.

PS: Utils needs to unify it's cases: I used camelCase, but some other functions use underscore cases.


Ostico:
Thank you, @SimZal

PR #397, opened 2016-09-10
Glossaries in combination with lexiQA

dfoteinakis:
The problem was the invocation of UI.clearMarks function before LexiQA did the highlighing.
I removed the calls within the lexiQA code.


Ostico:
Thank you @dfoteinakis ,

we tested and it works.

PR #405, opened 2016-09-14
adds html partial included in all pages

freegenie:
This is to set global variables site wide like the ones
we set to know if the user is logged in or the user email
and so on.

This also removes config.isAnonymousUser variable in favour of
isLoggedIn.

### DIFF
--- a/lib/Controller/AbstractControllers/viewController.php
+++ b/lib/Controller/AbstractControllers/viewController.php
@@ -52,9 +52,7 @@ abstract class viewController extends controller {
      *
      * @var Users_UserStruct
      */
-    protected $logged_user = false;
-
-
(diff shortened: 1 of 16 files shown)


freegenie:
@Ostico this is heavy, let's discuss it.

lib/Controller/AbstractControllers/viewController.php
```diff
@@ -52,9 +52,7 @@
      *
      * @var Users_UserStruct
      */
-    protected $logged_user = false;
-
-
+    protected $logged_user;
```
freegenie: @Ostico this was never false anyway because it was overridden in __construct with new UserStruct();

PR #515, opened 2016-11-17
Fix tag lock customize icon

riccio82:


riccio82:
[Stile Icona Tag Locks da rivedere](https://trello.com/c/e1uBTjEP/765-stile-icona-tag-locks-da-rivedere)

PR #610, opened 2017-01-19
lexiQA PartnerId configurable - suggestions update

dfoteinakis:
1) We would like to have LXQ_PARTNERID as a a configuration option. It will be provided by lexiQA along with the license key.

2) Update to lexiQA Suggestions on tooltip: Cursor set to pointer in order to show that the item is clickable

### DIFF
--- a/inc/INIT.php
+++ b/inc/INIT.php
@@ -118,9 +118,13 @@ class INIT {
      *
      */
     public static $LXQ_LICENSE = false;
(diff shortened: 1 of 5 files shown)


Ostico:
This new feature is now in test in the staging area.

dfoteinakis:
From what I see, everything is working properly.
Thank you

PR #681, opened 2017-03-07
Project queue

Ostico:
- Worker for create projects
- Store result data on Redis
- API to poll for the status of the queue
- Cyclic barrier to poll for project completion and get the response to pass the API ***new*** V1
- Cyclic barrier to poll for project completion via API V2 in the UI
- Completed UI Fixes
- API tested
- GDrive tested
- Bug fix in the API 'new' when a tmx was provided and no key is set
- Bug fix for converted files, hash position in the storage was wrong
- Removed an unuseful method in queries.php file \m/


Ostico:
Merged from other branches

PR #690, opened 2017-03-10
Matecat Multi Instance and Analysis parallelized

Ostico:
- ## Project Queue.

- ## MateCat can work in multiple instances ( namespaces ) on the same server.

Task Manager now can run in multiple instances:
- Same Host with different instances
- Different hosts with the same instance

Fast Analysis is also parallelized:
- Same host different instances
- Same host same instance ( multiprocess )


Ostico:
Merged from other branches

PR #703, opened 2017-03-27
Lexiqa: change https to http

riccio82:


Ostico:
They added a SSL certificate now.

PR #760, opened 2017-04-22
Fix Error when issue is hovered in sidepanel

dfoteinakis:
This patch fixes the issue when the selection starts, ends or contains a Matecat or lexiQA marked up tag.
It works by calculating the start/end offset of the selection on the base (clear from decorations) text and set both start/end node to 0. 
This needs to happen, because the text to be highlighted in the review side panel, does not contain any mark up, thus there are no nodes other than 0.


freegenie:
I pulled your changes and rebased on HotFix, I'm pushing a branch on our repo and close this PR. I'm closing this.

PR #777, opened 2017-05-10
Fix first_job_segment for chunks

riccio82:


riccio82:
[First_job_segment for chunks](https://trello.com/c/DUMOYPuj/439-first-job-segment-for-chunks)

PR #780, opened 2017-05-15
API to create projects in detatched mode

freegenie:
This creates a new /api/v1/new endpoint taking the same input params as /api/new, which responds immediately after queueing the project for creation. 

Some refactoring was needed to get projectName and analyzeURL immediately, while it was before delegated to project creation step which happens in background.

### DIFF
--- a/lib/Utils/ProjectManager.php
+++ b/lib/Utils/ProjectManager.php
@@ -158,6 +158,25 @@ class ProjectManager {
     }
 
     /**
+     * Project name is required to build the analyize URL. Project name is memoized in a instance variable
+     * so to perform the check only the first time on $projectStructure['project_name'].
+     *
+     * @return bool|mixed
+     */
+    protected function _sanitizeProjectName() {
+        $newName = self::_sanitizeName( $this->projectStructure[ 'project_name' ] );
+
+        if ( !$newName ) {
+            $this->projectStructure[ 'result' ][ 'errors' ][] = array(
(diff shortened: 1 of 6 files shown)


lib/Utils/ProjectManager.php
```diff
@@ -158,6 +160,32 @@ public function __construct( ArrayObject $projectStructure = null ) {
     }
 
     /**
+     * Project name is required to build the analyize URL. Project name is memoized in a instance variable
+     * so to perform the check only the first time on $projectStructure['project_name'].
+     *
+     * @return bool|mixed
+     */
+    protected function _getProjectName() {
```
freegenie: So the test to perform is that when name is not valid, UI should emit the same error as before, and APIs should emit the same error as well.

freegenie:
This new version uses a function to sanitize the project structure. Right now only the project name is sanitized. Given the projectStructure can be edited by objects outside the projectManager, it's better to sanitize the projectStructure right before projectCreation. 

Sanitization is called before the project is enqueued for creation so to have the project name cleaned up when providing the analyzeURL in newDetatchedController.

freegenie:
I was not able to trigger the error on  Invalid project name yet. Weird names are translated to underscore "_". 

We have a problem when this happens, because `Utils::friendly_slug` function returns "" when project_name is "_", so the resulting URL is broken. This problem is not related to this change though so I think this can be merged.

PR #785, opened 2017-05-22
Bug Fix:translations after a enter span disappear.

riccio82:
Fix: restore contenteditable=true for the span, like this is not possible write after return.


riccio82:
[[ BUG ] quando si clicca Rejected, la traduzione nel segmento sparisce](https://trello.com/c/gyCYiBTM/460-bug-quando-si-clicca-rejected-la-traduzione-nel-segmento-sparisce)

PR #809, opened 2017-06-17
lexiQA update

dfoteinakis:
Update includes:
- Use the new lexiQA API to create a project in lexiQA with the Matecat project TMX. This allows the creation of all segments at once and we avoid sending multiple requests when the page loads.
- Use only the matecat JobId as the project identifier. This allows to use the same QA data across jobs with changed password (avoids project recreation in lexiQA)
- Matecat fixes: Fixed an exception when translate (or review) loads and there is no initial segment selected

@freegenie This update requires a specific version of lexiQA, so it will not work in the production system. For this purpose you need to point the beta.matecat.com server to (https://)testbackend.lexiqa.net

### DIFF
--- a/public/js/cat_source/lxq.main.js
+++ b/public/js/cat_source/lxq.main.js
@@ -83,7 +83,7 @@ LXQ.init  = function () {
               licenseKey: config.lxq_license,
               partnerId: config.lxq_partnerid,
               lxqServer: config.lexiqaServer,
-              projectId: config.id_job+'-'+config.password
+              projectId: config.id_job
           }
       );
     }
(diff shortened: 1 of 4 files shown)


public/js/cat_source/lxq.main.js
```diff
@@ -1537,10 +1656,10 @@ LXQ.init  = function () {
                         results.qaurl = "#";
                     }
 
-                    if ( LXQ.enabled() ) {
-                        LXQ.doQAallSegments();
-                        //LXQ.refreshElements();
-                    }
+                    // if ( LXQ.enabled() ) {
```
freegenie: please don't leave such comments around

dfoteinakis:
Removed comments and dead code

PR #815, opened 2017-06-21
Fix outsource: event attached too many times

riccio82:


riccio82:
[Modale Outsource - Alcuni eventi vengono triggerati più volte](https://trello.com/c/yyFohHIs/626-modale-outsource-alcuni-eventi-vengono-triggerati-pi%C3%B9-volte)

PR #842, opened 2017-07-05
Fix: Somali is not RTL language

riccio82:


riccio82:
[Orientamento testo in somalo](https://trello.com/c/oZYVzrRb/635-orientamento-testo-in-somalo)

PR #961, opened 2017-09-18
fix improper usage of SQL IN for score recomputation

freegenie:


lib/Model/LQA/ChunkReviewDao.php
```diff
@@ -90,30 +90,29 @@ public static function getPenaltyPointsForChunk( \Chunks_ChunkStruct $chunk ) {
      * @return int
      */
     public static function getReviewedWordsCountForChunk( \Chunks_ChunkStruct $chunk ) {
+        $statuses = \Constants_TranslationStatus::$REVISION_STATUSES ;
```
Ostico: This is not an Array, array_merge will fail

Ostico: https://github.com/matecat/MateCat/pull/961/files?diff=split#diff-9fe637b96e0de527e8efe0e1e70a380fR108

PR #1067, opened 2017-12-12
fix @ at the beginning of translation

vincy:


lib/Utils/Engines/MyMemory.php
```diff
@@ -147,22 +188,22 @@ public function get( $_config ) {
...
 
-        $parameters                     = [];
-        $parameters[ 'seg' ]            = $_config[ 'segment' ];
-        $parameters[ 'tra' ]            = $_config[ 'translation' ];
-        $parameters[ 'tnote' ]          = $_config[ 'tnote' ];
-        $parameters[ 'langpair' ]       = $_config[ 'source' ] . "|" . $_config[ 'target' ];
-        $parameters[ 'de' ]             = $_config[ 'email' ];
-        $parameters[ 'prop' ]           = $_config[ 'prop' ];
+        $parameters               = [];
+        $parameters[ 'seg' ]      = ltrim( $_config[ 'segment' ] );
```
Ostico: `ltrim` in these rows lacks the characters to remove.

PR #1104, opened 2018-01-10
Added manageProjectType callback

vincy:


lib/Model/Segments/SegmentNoteDao.php
```diff
@@ -24,6 +24,18 @@ public static function bulkInsertFromProjectStructure( $notes ) {
 
             foreach ( $json_segment_ids as $id_segment ) {
                 foreach ( $json_entries as $json ) {
+                    $json_decoded = json_decode($json);
```
Ostico: Hi @vincy sorry, this PR can not be accepted.

This transformation must be done inside the Paypal Plugin.
This is the right place:
https://github.com/matecat/paypal_plugin/blob/master/lib/Features/Paypal/Utils/CDataHandler.php

Ostico: In this position MateCat must not perform any modification to the json, this is a simple insert without formatting.

PR #1131, opened 2018-02-12
Outsource trans info

riccio82:


riccio82:
[Outsource - Cambiare loader outsource](https://trello.com/c/6GgEFSsK/916-outsource-cambiare-loader-outsource)

PR #1132, opened 2018-02-12
add user id and user email in edit log CSV

freegenie:
@Ostico there you go 🍲 .


freegenie:
I've fixed the cache on UserDao query to one month.

PR #1168, opened 2018-03-19
documentation cancel, archive and active api

vincy:


lib/Plugins/Features/SegmentFilter/Model/SegmentFilterDao.php
```diff
@@ -109,62 +161,94 @@ private static function __getLimit( Chunks_ChunkStruct $chunk, FilterDefinition
...
+        ];
     }
 
     public static function findSegmentIdsForSample( Chunks_ChunkStruct $chunk, FilterDefinition $filter ) {
 
-        $limit = self::__getLimit($chunk, $filter);
+        if ( $filter->sampleSize() > 0 ) {
+            $limit = self::__getLimit( $chunk, $filter );
+        } else {
+            $limit = (object)[ 'limit' => 0, 'count' => 0, 'sample_size' => 0 ];
```
Ostico: You created an empty object for the `$limit` , but `NULL` is not enough? 
`$filter->sampleSize()` by definition can not be null or 0 ( you wrote an if for that ) when a `$filter->sampleType()` is one of them that use the `$limit` object. 
I'm wrong?

lib/Plugins/Features/SegmentFilter/Model/SegmentFilterDao.php
```diff
@@ -252,9 +340,133 @@ public static function getSqlForSegmentLength( $limit, $where, $sort ) {
...
+           $where->sql
+           ORDER BY st.id_segment
+        ";
+
+        return $sql;
+    }
+
+    public static function getSqlForToDo( $where ) {
+
+        $sql = "
```
Ostico: This query should accept TRANSLATED status when in revision

lib/Plugins/Features/SegmentFilter/Model/SegmentFilterDao.php
```diff
@@ -252,9 +340,133 @@ public static function getSqlForSegmentLength( $limit, $where, $sort ) {
...
+           $where->sql
+           ORDER BY st.id_segment
+        ";
+
+        return $sql;
+    }
+
+    public static function getSqlForRepetitions( $where ) {
+
+        $sql = "
```
Ostico: I see 3 equal queries here:
`getSqlForRepetitions`
`getSqlForFuzzies`
`getSqlForMT`

PR #1223, opened 2018-05-07
Fix: revise track changes

riccio82:


riccio82:
[Diff view in revise non funziona](https://trello.com/c/z70magMZ/988-diff-view-in-revise-non-funziona)

PR #1227, opened 2018-05-10
Bugs Fix

riccio82:


riccio82:
[Rinominare Concordance con TM Search](https://trello.com/c/dFhzP0D6/1007-rinominare-concordance-con-tm-search)

PR #1269, opened 2018-06-12
 get projects by team and name

vincy:


lib/Controller/API/V2/TeamsProjectsController.php
```diff
@@ -58,12 +59,40 @@ protected function _appendSingleProjectTeamValidators(){
         return $this;
     }
 
+    protected function _appendProjectTeamValidators($project){
```
Ostico: @vincy if this is not used we can remove?

vincy:
Yes, sure. I'm going to remove it.

PR #1279, opened 2018-06-19
Hot fix: lexiqa warning breaks segments

riccio82:


riccio82:
[Click su T++ fa sparire i segmenti](https://trello.com/c/H9LTzO2r/1058-click-su-t-fa-sparire-i-segmenti)

PR #1280, opened 2018-06-19
Tilde MT selector webcomponent integration

### LINKED ISSUE #1278: Tilde MT provider configuration options are broken
Here's a screenshot

image

Everything's misaligned and `From` and `To` dropdowns are empty.

I'll follow up with a PR with proposed fixes.

rihardsk:
To update the current UI and to better guard it from potentially breaking in the future I would like to refactor the Tilde MT plugin to use a separate [web component](https://en.wikipedia.org/wiki/Web_Components) for it's user-facing configuration dialogues. This fixes #1278 

The configuration UI is now fully implemented by a `tildemt-selector` web component. It now handles loading and displaying the Tilde MT system list and available term corpora. This is done by calling the `letsmt.eu` API directly from JavaScript. The configuration end-result is a JSON object in the form
```JSON
{
  "engineName": "My Tilde MT engine for en-lv and en-et",
  "client-id": "u-745522aafd-efeb-5tadda-a6a9c-eb844369df7179",
  "systems": {
    "en-lv": {
      "params": {
        "systemID": "smt-89ce23df-462c-4910-a5fe-556a9rdd34caad",
        "options": "client=localhost,version=1.0,termCorpusId=c-40b1e2d5-e752-4d05-8691-aa56d562afa"
      }
    },
    "en-et": {
      "params": {
        "systemID": "smt-73t895d5cc-6bee-494a-a681-1cf759a9956c",
        "options": "client=localhost,version=1.0"
      }
    }
  }
}
```
It is passed to the back-end which can then directly use the `params` object to make translation requests for that language pair (and the `client-id` value must be added in request headers). 

As a result the back-end code is now simplified. Since it no longer needs to handle system and term list loading a lot of stuff can be thrown out.

We're hosting the web component at [letsmt.eu/ws/tildemt-selector.html](https://www.letsmt.eu/ws/tildemt-selector.html). This is to be able to deploy non-breaking feature updates and bug-fixes when we see fit in the future.

One thing that came up when testing the integration is that we weren't able to receive accepted post-edits (the `set()` method just wasn't being called). We verified that this is also the case for the current implementation in Matecat.com. Are we doing something wrong or has this feature been cut?

Here are a couple of screenshots from within the UI. If you need a Client ID for testing I can send it via email.
image
image


Ostico:
Hi @rihardsk ,
regarding the client ID, of course we'd like to have a testing account, so if you have one and you can send it to me via email it would be great. We will make tests on our side also.

Regarding web component, my only concern is that if we change the interface to add react componets or changing the style of the page, and in future we will do, the style of the component and the style of the matecat will differ.

Regarding the `set()` method, sorry but i need more info because i don't realized of what method we are talking.

rihardsk:
Hi @Ostico ,
I've sent you a client ID to an email i found associated with one of your commits.

Regarding changing the style of the web component, I understand your concern. This has also been a thing I've pondered about when integrating the web component into MateCat and trying to match it's style to MateCat's. The solution that I opted for was [injecting custom styling via JS](https://github.com/matecat/MateCat/pull/1280/files#diff-5f35cef7bcdffe17e7fbc27c5829a6e1R150). The solution is less than ideal but I see no other choice until CSS mixins are supported in browsers. I think here the benefits of it being a web component outweigh the negative aspects of the current hacky solution for adjusting the style.

I'm not familiar with react or react components that well so I don't know what the conflicts there might be. Could you expand on that?

By `set()` I mean [this method here](https://github.com/matecat/MateCat/blob/8c1c1f6f1336743283bbc67b504026de86be2da2/lib/Utils/Engines/LetsMT.php#L229). It used to be that the Tilde MT (LetsMT) plugin would receive accepted post-edited segments so that translations could be improved for MT systems that support that. It seems that either something has broken or I'm doing something wrong when trying to invoke the feature.

riccio82:
Hi @rihardsk ,

For now, the solution can be fine. When we will make the refactoring of that part in React each MT could have its own React component with its custom css and configurations.

About the `set` method i will check what happens.

riccio82:
You're right, we've introduced an error where sets are not made on the MTs, we'll fix it as soon as possible.

PR #1287, opened 2018-06-25
drive doc mime

vincy:


lib/Utils/ConnectedServices/GDrive/Session.php
```diff
@@ -434,63 +434,44 @@ public function importFile( $fileId ) {
...
-            } else {
-                $downloadUrl = $file->getDownloadUrl();
+            $file = $service->files->get($fileId);
+            $export_mime = RemoteFileService::officeMimeFromGoogle($file->mimeType);
+            if ($export_mime == RemoteFileService::MIME_DOCX) {
+                $export_mime = RemoteFileService::MIME_RTF;
             }
 
-            if ($downloadUrl) {
+            $content = $service->files->export($fileId, $export_mime)->getBody()->__toString();
```
Ostico: You tried if this works with other formats ( spreadsheet and pptx)?

vincy: Yes, I tried using spreadsheet and it works. Do you want me to try pptx as well?

Ostico: If possibile, try all the cases.

PR #1293, opened 2018-07-05
public api team

vincy:


lib/Controller/API/App/TeamPublicMembersController.php
```diff
@@ -0,0 +1,38 @@
...
+class TeamPublicMembersController extends KleinController {
+
+    /**
+     * Get team members list
+     */
+    public function publicList(){
+
+        $team = ( new TeamDao() )->setCacheTTL( 60 * 60 * 24 )->findById( $this->request->id_team );
+        $teamModel = new TeamModel( $team );
+        $teamModel->updateMembersProjectsCount();
```
Ostico: This is not needed, because this is an expensive method to get the member list.

Lets call directly Membership Dao. And expand the users info to pass to the renderer.

PR #1298, opened 2018-07-11
Fix header in safari

KromPass:


Ostico:
@rihardsk , sorry we removed the we component because this conflicts in some manner with our event listeners. The console was reporting an error in one of our libraries making this hard to debug and find. We spent approximately 6 hours to find the root of the problem and to understand that your component was causing the issue.
As result, the header was broken in Safari.

riccio82:
We have not figured out where is the fault.You can test it with Safari, adding this lines in the upload.html page (line 46)  there are errors in the console and the page is not displayed correctly (the header). 

`<!-- add web component support for older browsers -->`
`<SCRIPT src="https://cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/0.7.22/webcomponents.js"></SCRIPT>`
``
`<link id="tildemt_link" rel="import" href="https://www.letsmt.eu/ws/tildemt-selector.html"/>`

If we add the same code before the script `<script type="text/javascript" src="${basepath}public/js/build/libs_upload.js?build=${build_number}"></script>` It seems to work.

I would like to understand why the header fails to know if it can give some other type of problem.

PR #1328, opened 2018-08-21
Add  iADAAPTA connector 

e-gomez:
new plugin to be integrated inside MateCat and it calls to the iADAAPTA project to obtain the translations.


e-gomez:
IADAATPA has changed to MT-HUB

PR #1386, opened 2018-11-17
in case of fast or tm analysis shutdown.

tyraelll:
-Bash script to check fast and tm analysis => for cron purpose. 
-Created in need "down of tm analysis.".
-You need to assign cronjob for everyminute.

If anyone of them gone, you cant get analysis on occasion of file upload( endless queue loop ). So i've created this script to be sure screens are always live.


Ostico:
Thank you, as improvement, we suggest you to create the ` /etc/init.d/ ` scripts and put them into Monit to check the analysis status.

PR #1406, opened 2019-01-16
Decouple review extended

freegenie:
Remove ReviewImproved namespace from ReviewExtended and classes dependencies swapped.


freegenie:
rebased onto develop.

PR #1408, opened 2019-01-16
fix missing update of reviewed_words_count on bulk status update

freegenie:


lib/Model/Translations/BulkStatusChangeModel.php
```diff
@@ -0,0 +1,87 @@
...
+            $this->unchangeble_segments = Translations_SegmentTranslationDao::getUnchangebleStatus( $this->segment_ids, $this->destination_status  );
+            $this->changeable_segments = array_diff( $this->segment_ids, $this->unchangeble_segments  );
+        }
+        return !empty( $this->changeable_segments ) ;
+    }
+
+    protected function _statusIsValid() {
+        return in_array(
+                $this->destination_status, [
+                        Constants_TranslationStatus::STATUS_TRANSLATED,
```
Ostico: Remove Approved status

lib/Controller/API/V2/JobStatusController.php
```diff
@@ -37,26 +38,27 @@ protected function afterConstruct() {
...
-                }
-                catch(\Exception $e){
-                    $this->response->json( [ 'data' => true, 'unchangeble_segments' => $segments_id] );
-                    return;
-                }
-            }
 
-            $this->response->json( [ 'data' => true, 'unchangeble_segments' => $unchangeble_segments, 'stats' => $stats ] );
+        $bulk_status_change = new BulkStatusChangeModel( $this->job, $segments_id ) ;
+        if ( $bulk_status_change->anyChangeableSegmentStatus() ) {
```
Ostico: Pass the requested status to the method or class.

lib/Model/Translations/SegmentTranslationDao.php
```diff
@@ -274,6 +274,27 @@ public static function getSegmentsWithIssues( $job_id, $segments_ids ) {
         return $stmt->fetchAll();
     }
 
+    public static function getCountForReviwedWordsBySegmentId( $job_id, $segment_ids ) {
+        $sql = " SELECT SUM( raw_word_count ) FROM segments s
+            JOIN segment_translations st ON st.id_segment = s.id
+            WHERE st.id_job = ?
+           AND match_type != 'ICE'
```
Ostico: Remove and change misleading method name

lib/Model/Translations/BulkStatusChangeModel.php
freegenie: Both Translated and Approved statuses are to be considered valid indeed, otherwise the we won't be able to add or subtract reviewed_words_count.

lib/Controller/API/V2/JobStatusController.php
freegenie: done

freegenie:
Closing this pull request because as we said we need to process the bulk status change as background process.

PR #1421, opened 2019-02-22
Hot fix

rubensantillan:


rubensantillan:
[Correggere mail MateCat - Resource shared](https://trello.com/c/kSV2Mj2k/1599-correggere-mail-matecat-resource-shared)

PR #1429, opened 2019-02-28
Fix upload zip from api

riccio82:


Ostico:
Not valid, if the token is empty no validation is performed. The meaning of this row is: if there is not an agreed token ( cookie or server generated ) the upload **must** fail

PR #1488, opened 2019-05-10
Revert "Fix: tag menu with glossary tags"

Ostico:
Reverts matecat/MateCat#1487

This change crash when there are no glossary marks but there are tags


Ostico:
@riccio82 Watch on this WARNING

public/js/cat_source/es6/react/components/segments/TagsMenu.js
```diff
@@ -82,11 +82,7 @@ class TagsMenu extends React.Component {
         var sourceClone = $( '.source', UI.currentSegment ).clone();
         //Remove inside-attribute for ph with equiv-text tags
         sourceClone.find('.locked.inside-attribute').remove();
-        sourceClone.html(sourceClone.find('mark'));
```
Ostico: This line deletes the `html()` value of `sourceClone` if there are no `'mark'`

PR #1508, opened 2019-06-05
Word count raw: links as 1 word

Ostico:


Ostico:
https://trello.com/c/KBFcz5lT/1757-matecat-fix-for-double-encoded-non-breaking-space-in-raw-word-count-ampnbsp-count-as-1-word

PR #1548, opened 2019-08-12
[S3 file storage] File metadata

mauretto78:
Added a file metadata object, used to manage queue.

Migration file is included (two fields are added: 'metadata' and 'is_converted').

### DIFF
--- a/migrations/20190812091652_file_meta.php
+++ b/migrations/20190812091652_file_meta.php
@@ -1,0 +1,24 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class FileMeta extends AbstractMigration
+{
+    public $sql_up = <<<EOF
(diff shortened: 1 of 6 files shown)


migrations/20190812091652_file_meta.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class FileMeta extends AbstractMigration
+{
+    public $sql_up = <<<EOF
+      ALTER TABLE `files` ADD COLUMN `is_converted` TINYINT NULL AFTER `sha1_original_file`;
```
Ostico: When possible, place fields in tail position of the table and use non locking algorithm:
```
ALTER TABLE `files` ADD COLUMN `is_converted` TINYINT NULL , ALGORITHM = INPLACE, LOCK=NONE ;
```

PR #1602, opened 2019-12-17
FindElementInS3CacheTask update

mauretto78:


lib/Plugins/Features/AbstractRevisionFeature.php
```diff
@@ -66,21 +66,25 @@ public function filterContributionStructOnSetTranslation( ContributionSetStruct
...
+     * @param int $id_job
+     * @param int $source_page
+     *
+     * @throws NotFoundException
      */
-    public function filter_review_password_to_job_password( $review_password, $id_job ) {
-        $chunk_review = ChunkReviewDao::findByReviewPasswordAndJobId(
-                $review_password, $id_job );
+    public function filter_review_password_to_job_password( $review_password, $id_job, $source_page ) {
+        $chunk_review = (new \LQA\ChunkReviewDao())->findByJobIdReviewPasswordAndSourcePage( $id_job, $review_password, $source_page );
```
Ostico: Return $chunk_review instead of password ( check where this method is used and change accordingly ) because it's not usefult to make a query to get a data we already fetched here.

lib/Controller/catController.php
```diff
@@ -285,6 +255,34 @@ public function doAction() {
...
+                    'filter_review_password_to_job_password',
+                    $this->password,
+                    $this->jid,
+                    Utils::getSourcePage()
+            );
+        }
+
+        $this->chunk = Chunks_ChunkDao::getByIdAndPassword( $this->jid, $this->password );
+
+        $this->featureSet->run('catControllerChunkFound', $this);
```
Ostico: Check this callback i think can be simplyfied/removed, it seems to do the same think the previous filter/callback does.

PR #1628, opened 2020-01-29
added outsource disablement by env var

aijanai:
This allows to set an env var ENABLE_OUTSOURCE="false" and disable outsource quotation in volume analysis at runtime.

When ENABLE_OUTSOURCE is not defined, outsource is enabled as usual.


Ostico:
Good

PR #1663, opened 2020-03-23
HotFix: Fix split job - Migration required

mauretto78:
Full working version, last commit before PR & code review

### DIFF
--- a/lib/Utils/ProjectManager.php
+++ b/lib/Utils/ProjectManager.php
@@ -1462,8 +1462,6 @@ class ProjectManager {
 
         $reverse_count = [ 'eq_word_count' => 0, 'raw_word_count' => 0 ];
 
-        $standard_analysis_wc = round( $projectStructure[ 'standard_analysis_wc' ] / 2 );
-
         foreach ( $rows as $row ) {
 
             if ( !array_key_exists( $chunk, $counter ) ) {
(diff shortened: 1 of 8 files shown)


lib/Utils/ProjectManager.php
```diff
@@ -1598,22 +1615,22 @@ protected function _splitJob( ArrayObject $projectStructure ) {
             /**
              * Async worker to re-count avg-PEE and total-TTE for splitted jobs
              */
-            SplitQueue::recount( $jobInfo );
+            SplitQueue::recount( $jobToSplit );
```
Ostico: Check what happens here.

migrations/20200323145227_add_standard_wc_to_job_table.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class AddStandardWcToJobTable extends AbstractMigration  {
```
Ostico: This migration doesn't works because you are extending AbstractMigration.
Please change this and extend AbstractMatecatMigration.

Take a look to other migration files.

mauretto78: The migration file was modified as you suggested.

lib/Utils/ProjectManager.php
mauretto78: This async worker recounts avg-PEE and total-TTE for splitted jobs.

The endpoint of this process is **JobsWorker::_recountAvgPee()**; inside of that all segments are looped and for every segment is calculated the weighted PEE and the TEE.

Jobs_JobDao::updateJobWeightedPeeAndTTE() is called at the end.

This method executes this query:

```
UPDATE jobs 
                    SET avg_post_editing_effort = :avg_post_editing_effort, 
                        total_time_to_edit = :total_time_to_edit 
                    WHERE id = :id 
                    AND password = :password
```

Do you have enough information or do you need more? Let me know

Ostico: thanks this is enough :)

Ostico:
Merged in develop

PR #1666, opened 2020-03-24
Paginated api team

mauretto78:
V2 API Team Projects is now fully paginated.

Two parameters are allowed:

page (default value = 1)
step (default value = 20)

Example:

https://xxx.matecat.com/api/v2/teams/1/projects?page=10&step=3

If no parameter is provided, page and step are set to corresponding default values.

### DIFF
--- a/lib/Controller/API/V2/TeamsProjectsController.php
+++ b/lib/Controller/API/V2/TeamsProjectsController.php
@@ -96,34 +96,89 @@ class TeamsProjectsController extends KleinController {
         $this->response->json( [ 'projects' => $projects ] );
     }
 
+    /**
+     * @throws NotFoundException
+     * @throws \Exceptions\NotFoundException
+     * @throws \Exception
+     */
     public function getAll() {
 
+        $id_team = $this->request->param( 'id_team' );
+        $page    = $this->request->param( 'page' ) ? $this->request->param( 'page' ) : 1;
+        $step    = $this->request->param( 'step' ) ? $this->request->param( 'step' ) : 20;
+
         $this->featureSet->loadFromUserEmail( $this->user->email );
-
-        $projectsList = \Projects_ProjectDao::findByTeamId( $this->params[ 'id_team' ], 60 );
-
-        $https = $this->request->server()->get('HTTPS');
-        $baseUrl = isset($https) && $https != 'off' ? 'https' : 'http' . $this->request->server()->get('SERVER_NAME');
-
-        $_links      = [
-                "base" => $baseUrl,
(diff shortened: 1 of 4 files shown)


lib/Model/Projects/ProjectDao.php
```diff
@@ -148,19 +148,29 @@ public function massiveSelfAssignment( TeamStruct $team, Users_UserStruct $user,
...
      * @param int $ttl
      *
      * @return DataAccess_IDaoStruct[]
      */
-    public static function findByTeamId( $id_team, $ttl = 0 ) {
+    public static function findByTeamId( $id_team, $limit, $offset, $ttl = 0 ) {
         $thisDao = new self();
         $conn    = Database::obtain()->getConnection();
-        $stmt    = $conn->prepare( self::$_sql_get_projects_for_team );
+        $stmt    = $conn->prepare( self::$_sql_get_projects_for_team . " LIMIT " . $limit . " OFFSET " . $offset );
```
Ostico: Those parameters do no pass through prepared statement, do not concatenate them without an explicit cast to int

lib/Controller/API/V2/TeamsProjectsController.php
```diff
@@ -89,25 +89,99 @@ public function getByName() {
...
+        $step    = $this->request->param( 'step' ) ? $this->request->param( 'step' ) : 20;
+
+        $this->featureSet->loadFromUserEmail( $this->user->email );
+        $projectsList = \Projects_ProjectDao::findByTeamId( $id_team, $step, $this->getOffset($page, $step) ,60 );
 
-        $this->featureSet->loadFromUserEmail( $this->user->email ) ;
+        $projectsList = ( new Project( $projectsList ) )->render();
 
-        $projectsList = \Projects_ProjectDao::findByTeamId( $this->params[ 'id_team' ], 60 );
+        $totals      = \Projects_ProjectDao::getTotalCountByTeamId( $id_team, 0 );
```
Ostico: Maybe we should try to cache this request for 60 seconds.

lib/Model/Projects/ProjectDao.php
mauretto78: I tried to create a full prepared statement, but _fetchObject method returns me a syntax error.
I cast $limit and $offset to int as you suggested, hope this wuold be enough.

lib/Controller/API/V2/TeamsProjectsController.php
mauretto78: Cache has been set to 60 seconds.

PR #1670, opened 2020-03-27
Glossary async

mauretto78:


inc/task_manager_config.ini.sample
```diff
@@ -162,4 +162,31 @@ ALIGNER_SEGMENT_CREATE[queue_name]    = "aligner_segment_create"
...
 ALIGNER_SEGMENT_CREATE[max_executors] = 5
 ALIGNER_SEGMENT_CREATE[redis_key]     = "aligner_segment_create_list"
-ALIGNER_SEGMENT_CREATE[loggerName]    = "aligner_segment_create.log"
\ No newline at end of file
+ALIGNER_SEGMENT_CREATE[loggerName]    = "aligner_segment_create.log"
+
+;;##############################
+;;##### Key Name Glossary  #####
+;;##############################
+GLOSSARY[queue_name]    = "glossary"
```
Ostico, line 170: @mauretto78 what is this queue, please can you explain?

Ostico: Can you remove this from the pr? thanks

inc/task_manager_config.ini.sample
```diff
@@ -162,4 +162,31 @@ ALIGNER_SEGMENT_CREATE[queue_name]    = "aligner_segment_create"
...
+GLOSSARY[pid_set_name]  = "ch_pid_glossary"
+GLOSSARY[max_executors] = 1
+GLOSSARY[redis_key]     = "glossary"
+GLOSSARY[loggerName]    = "glossary.log"
+
+;;##################################
+;;##### Key Name GlossaryRead  #####
+;;##################################
+GLOSSARY[queue_name]    = "glossary_read"
```
Ostico, line 179: I think that there is a mistake on the two GLOSSARY ini key. With the same name "GLOSSARY" and key "queue_name" the second block will override the first one.

They should be called like the names you defined in the controller:
https://github.com/matecat/MateCat/pull/1670/files#diff-ac35daf33d7d4a73748023e5fcedaf05R13

mauretto78:
Done, I removed GLOSSARY block and renamed GLOSSARY_WRITE and GLOSSARY_READ keys.

PR #1726, opened 2020-05-27
Fix for CJK

mauretto78:


lib/Utils/QA.php
```diff
@@ -1531,12 +1531,43 @@ protected function _checkTagsBoundary() {
         //so, if we found a last char mismatch, and if it is in the source: add to the target else trim it
         if ( ( count( $source_tags[ 0 ] ) != count( $target_tags[ 0 ] ) ) && !empty( $source_tags[ 0 ] ) || $source_tags[ 1 ] != $target_tags[ 1 ] ) {
 
-            //Append a space to target for normalization.
-            $this->target_seg = rtrim( $this->target_seg );
-            $this->target_seg .= $source_tags[ 1 ][ 0 ];
+            // CJK = chinese, japanese, korean
+            $cjk = [
```
Ostico, line 1535: You can use `CatUtils::isCJK()` function here

mauretto78: As you requested:

https://github.com/mauretto78/MateCat/commit/f3cc302d845c6b5f7b9d8b83331b41cdd5bf534b

PR #1738, opened 2020-06-09
Detect MIME Type improvement

mauretto78:
1) Detect MIME Type improvement using File Information PECL library;
2) Restored 'text/x-fortran' as allowed extension (because of mime type false positives);

Sometimes the detection of mime type with PHP gives false positives; for example, a simple txt file with lines that begin with a single C letter plus spaces can be encoded as 'text/x-fortran', because this letter combination seems to be a Fortran style comment.

For more information see here:

https://stackoverflow.com/questions/16190929/detecting-a-mime-type-fails-in-php

### DIFF
--- a/lib/Utils/fileupload/upload.class.php
+++ b/lib/Utils/fileupload/upload.class.php
@@ -316,16 +316,16 @@ class UploadHandler {
      * @return mixed|string
      */
     private function getMimeContentType( $filename ) {
+        if ( function_exists( 'mime_content_type' ) ) {
+            return mime_content_type( $filename );
+        }
+
         if ( function_exists( 'finfo_open' ) ) {
             $finfo     = finfo_open( FILEINFO_MIME_TYPE );
             $finfoFile = finfo_file( $finfo, $filename );
             finfo_close( $finfo );
 
             return $finfoFile;
-        }
-
-        if ( function_exists( 'mime_content_type' ) ) {
-            return mime_content_type( $filename );
         }
 
         return 'application/octet-stream';
--- a/lib/Utils/fileupload/upload.class.php
+++ b/lib/Utils/fileupload/upload.class.php
@@ -232,8 +232,8 @@ class UploadHandler {
         $file->name     = $this->trim_file_name( $name );
         $file->size     = intval( $size );
         $file->tmp_name = $uploaded_file;
-        //$file->type = $type; // Override and ignore the client type definition
-        $file->type = mime_content_type( $file->tmp_name );
+
+        $file->type = $this->getMimeContentType( $file->tmp_name );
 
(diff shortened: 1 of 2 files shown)


lib/Utils/fileupload/upload.class.php
```diff
@@ -298,6 +298,39 @@ protected function handle_file_upload( $uploaded_file, $name, $size, $type, $err
...
+     *
+     * Even though some false positive are always possible. Take a look here:
+     *
+     * https://stackoverflow.com/questions/16190929/detecting-a-mime-type-fails-in-php
+     *
+     * @param $filename
+     *
+     * @return mixed|string
+     */
+    private function getMimeContentType( $filename ) {
```
Ostico, line 318: This is weird and I think is not useful since `finfo_file` and `mime_content_type` use the same primitives and they belongs to the same package: [ref.fileinfo](https://www.php.net/manual/en/ref.fileinfo.php) . 
`mime_content_type` is only a syntactic sugar for `finfo_file` [fileinfo.c](https://github.com/php/php-src/blob/PHP-5.6.40/ext/fileinfo/fileinfo.c#L601)
They are both based on the information from the `magic.mime` file.
In fact, from my tests they give the same results.

```
php > var_dump( mime_content_type( "./tempTestFile" ) );
string(14) "text/x-fortran"

php > $finfo     = finfo_open( FILEINFO_MIME_TYPE );
php > var_dump( finfo_file( $finfo, "./tempTestFile" ) );
string(14) "text/x-fortran"
```

Unix `file` function ( the one on Ubuntu Bionic ) instead detects a different and  right mime type:
```
#> file -bi tempTestFile
text/plain; charset=utf-8
```

mauretto78: You are right, the function `getMimeContentType()` in this case gave the same result ("text/x-fortran"). **So for this specific case there is no difference.** 

And yes, I checked also the file mime using `file` function in Linux (giving me the same results).

The aim of `getMimeContentType()` function was to have a safer approach (consider the case of mime_content_type returning false), maybe I can put the mime_content_type block on top.  

But if you prefer we can just leave the modified `Mime2Extension.php` file.

mauretto78: @Ostico I inverted the function call order inside `getMimeContentType()` function.

The function now does the same as before but in case it mime_content_type function is not active il will try to use finfo.

In case of failure an error message is added.

Ostico: I think it is superfluous since they belongs to the same package, so if there is no `mime_content_type` also `finfo_file`will not exists.

It's better to have a `shell_exec` on `file -b --mime-type fileName` as fallback.

PR #1754, opened 2020-06-19
Develop trailing spaces during analysis

mauretto78:


lib/Utils/Analysis/Workers/TMAnalysisWorker.php
```diff
@@ -160,6 +160,10 @@ protected function _updateRecord( QueueElement $queueElement ) {
         $eq_words       = $equivalentWordMapping[ $new_match_type ] * $queueElement->params->raw_word_count / 100;
         $standard_words = $eq_words;
 
+        // Init \PostProcess
+        $check = new \PostProcess( $queueElement->params->segment, $suggestion );
```
Ostico, line 164: Why those initializations are outside the if code block?
Moving the PostProcess definition outside the `if` code we are not performing any check on the MT return values.

Ostico: This is an error.

mauretto78: As you know the aim of this commit is to add QA->consistencyCheck() during TM analysis, in order to add trailing spaces as setTranslation actually does.

Indeed $this->_matches[ 0 ][ 'raw_segment' ] does not contain any trailing spaces, so perform consistencyCheck() on that as source is totally useless.

I moved initialization outside the if/else block because I used in both cases $queueElement->params->segment as source of \PostProcess.

Ostico: Yes, right, but the initialization of the PostProcess is not performed on the results of the MT `$this->_matches[ 0 ][ 'raw_segment' ]` so in this way we are not performing in the if block the MT any space realignment.

Ostico: It is true that `$suggestion == $queueElement->params->segment` but could not be in every cases, for example when some custom engine do something on the source.

mauretto78: I restored the if/else block as you requested, and added the block code which performs consistencyCheck after that.

So there are two \PostProcess invokes. (Maybe I should merge $err_json obtained from these two analysys)

What do you think about it?

PR #1777, opened 2020-07-24
Add install script v2.

YourLittleHelper:
- Add configuration parameters for SMTP server

- Add configuration parameters for Google OAuth

- Fix Machine Translation suggestion error

- Clean up systemd service definitions


Ostico:
Merged and moved in the support_scripts  directory.
INSTALL directory is not protected by .htaccess and we removed executable permissions on the file.

PR #1778, opened 2020-07-28
SignupController::doForgotPassword() complete refactoring

mauretto78:


mauretto78:
Improved forgotPassword behavior.

Now the controller responds with a JSON like that:

```
{
    "email": "mauro@translated.net",
    "wanted_url": "http://www.google.it",
    "errors": []
}
```

As you can see `errors` is an array. If it's empty the email was correctly enqueued.

PR #1822, opened 2020-10-20
Fix email referer

mauretto78:


Ostico:
Now looks cool :smile:

mauretto78:
@riccio82 

The back-end now is capable to build the complete URL of a job from job id/password combination, and segment id and revision number as optional parameters.

So, the UI should send an additional field called `revision_number` when sending the ajax request to persist a note (as it does for issues).

Possible values are:

- `null` for translation state
- `1` for R1 state
- `2` for R2 state

Thank you :)

PR #1843, opened 2020-11-04
setCurrentSegmentController fix

mauretto78:
Fix missing $nextSegmentId

### DIFF
--- a/lib/Controller/AbstractControllers/viewController.php
+++ b/lib/Controller/AbstractControllers/viewController.php
@@ -230,18 +230,25 @@ abstract class viewController extends controller {
      * @return bool|null
      */
     private static function getIsRevisionFromIdJobAndPassword(){
+
         $jid = static::getInstance()->jid;
         $receivedPassword = static::getInstance()->received_password;
 
-        $job = Jobs_JobDao::getByIdAndPassword($jid, $receivedPassword);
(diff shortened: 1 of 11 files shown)


lib/Controller/AbstractControllers/viewController.php
```diff
@@ -225,11 +216,53 @@ public function getAuthUrl() {
...
+            $isRevision = self::getIsRevisionFromRequestUri();
+        }
+
+        return $isRevision;
+    }
+
+    /**
+     * @return bool|null
+     */
+    private static function getIsRevisionFromIdJobAndPassword(){
```
Ostico, line 232: This method is similar to other methods ( Ex: the ChunkPasswordValidator ), can we check if it is possible to use a common method or validator? Anyway if not possible, add a cache ttl on the JobDao request and implement the same cache method on the `ChunkReviewDao::findByReviewPasswordAndJobId`

Ostico (changes requested):

lib/Controller/AbstractControllers/viewController.php
mauretto78: I created a new method inside `ChunkReviewDao` called `isTOrR1OrR2` which performs a tailored query for this purpose.

PR #1864, opened 2020-12-03
Whole text finder 1013

mauretto78:


mauretto78:
This includes also update to xliff parser v1.0.39

PR #1951, opened 2021-04-07
:art: Enable prettier for support_scripts folder

cloud-walker:
## Contents
- ignore package-lock.json files by prettier (they are generated automatically by npm)
- remove `support_scripts` folder from the prettierignore

### DIFF
--- a/support_scripts/tasks/model.json
+++ b/support_scripts/tasks/model.json
@@ -1,138 +1,184 @@
 {
-	"model": {
-        "version" : 1,
+  "model": {
+    "version": 1,
 
-		"label": "Ebay 1",
-		"categories": [{
-			"label": "Accuracy",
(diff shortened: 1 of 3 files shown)


support_scripts/tasks/model.json
```diff
@@ -1,138 +1,184 @@
 {
-	"model": {
```
cloud-walker, line 2: I think this file is not related to frontend stuff, but I personally don't think nothing bad to leave prettier doing its job here also 🤔 

@riccio82 what do you think? I can always ignore this folder if needed

riccio82: Yes, it is a json configuration file.
It's ok to check it.

riccio82 (approved):

PR #1952, opened 2021-04-07
:art: enable formatting for public folder files

cloud-walker:
## Contents
- improve structure of the .prettierignore file
- enable prettier for public/css folder
- enable prettier for public/api/swagger-source.js file
- enable prettier for public/img folder
- keep on prettierignore uglified / minified files (but tracked in the repo)

## Notes
Before reviewing, please hide whitespace changes:
image

The diff will go down from:
image
to
image

## Todos
- [x] Manually check every affected files

### DIFF
--- a/public/css/sass/style.scss
+++ b/public/css/sass/style.scss
@@ -1,128 +1,130 @@
 body {
-    transition: padding-top 200ms ease-out;
-    -webkit-transition: padding-top 200ms ease-out;
+  transition: padding-top 200ms ease-out;
+  -webkit-transition: padding-top 200ms ease-out;
 }
 
 body.incomingMsg {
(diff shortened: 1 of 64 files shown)


public/css/sass/style.scss
```diff
@@ -1,128 +1,130 @@
 body {
```
riccio82, line 1: Have you seen this file?
I still haven't found a way to figure out what's really used in here, it's really huge.

riccio82 (approved):

public/css/sass/style.scss
cloud-walker: Ahahahaah! One thing at a time, we will tackle this, eventually!

PR #1953, opened 2021-04-08
✅ First frontend test

cloud-walker:
## Draft PR
This PR is still a draft, before merge this please focus on the prettier related PRs first!

## Contents
- add basic jest configuration
- add `yarn test` script for testing in development
- add `yarn coverage` for analyzing code coverage and testing before deploy / during CI
- expose to global the lodash module (💩 )
- add first React component related test

Current coverage level:

image

Current InputField module coverage:

image

## Checks

- [ ] Verify that the `coverage/` folder generated by the `yarn coverage` script doesnt clash with others existing testing environments
- [x] ⚠️  `supports_scripts/grunt` package.json modules should be locked with `yarn.lock` as currently in development the package minors are not locked!


public/js/cat_source/es6/components/common/InputField.js
```diff
@@ -1,3 +1,4 @@
+import React from 'react'
```
cloud-walker, line 1: @riccio82 Here I've added the React import explicitly (shadowing the global React from the `react.js` bundle generated by the `Gruntfile.js`

Personally I think that it should not be a problem, as it should be the same React instance! Let me know if something is wrong!

riccio82: Yes it should be the same. It's ok for now.

riccio82 (approved):

PR #1971, opened 2021-04-20
Fix uber segments errors

mauretto78:


lib/Utils/QA.php
```diff
@@ -1314,7 +1303,7 @@ protected function _addThisElementToDomMap( DOMElement $element) {
             }
         }
 
-        return !(in_array($elementValue, $this->tagsToBeExcludedFromChecks) and $elementHasDataRef);
+        return !(in_array($elementValue, $tagsToBeExcludedFromChecks) and $elementHasDataRef);
```
Ostico, line 1306: We should prefer a cleaner syntax or a clear code comment to avoid brain burn when reading this line of code!

Ostico (changes requested):

lib/Utils/QA.php
mauretto78: I extracted the code in a more human-readable style:

https://github.com/matecat/MateCat/pull/1971/commits/61952d5b52092ba416e8efab7d46d10f8cef97da

Ostico (approved):

PR #1992, opened 2021-04-28
Add Travis file for CI

riccio82:


.travis.yml
```diff
@@ -0,0 +1,23 @@
+sudo: false
```
cloud-walker, line 1: This option is deprecated: https://config.travis-ci.com/ref/sudo

riccio82: Deleted

cloud-walker (approved):

PR #2010, opened 2021-05-06
Fixed step parameter and ghost segments

Ostico:


lib/Model/QualityReport/QualityReportSegmentModel.php
```diff
@@ -270,7 +270,7 @@ public function getSegmentsForQR( array $segment_ids, $isForUI = false ) {
             $seg->pee_translation_revise     = $seg->getPEEBwtTranslationRevise();
             $seg->pee_translation_suggestion = $seg->getPEEBwtTranslationSuggestion();
 
-            $segments[$i] = $seg;
```
Ostico, line 273: This was the cause of ghost segments. ` $i ` variable was used inside a nested loop.

mauretto78 (approved):

PR #2047, opened 2021-05-28
Fix whole text search

mauretto78:


lib/Controller/getSearchController.php
```diff
@@ -3,10 +3,10 @@
 use Features\ReviewExtended\ReviewUtils;
 use Features\TranslationVersions;
 use Matecat\Finder\WholeTextFinder;
+use Matecat\SubFiltering\MateCatFilter;
```
Ostico, line 6: @mauretto78 are you sure that we have to introduce the new namespace and classes here?

Ostico (changes requested):

PR #2064, opened 2021-06-10
Added Tamil India to supported languages

mauretto78:


Ostico (changes requested):
Change the label for "Tamil" in "Tamil India"

Ostico (approved):

PR #2140, opened 2021-07-30
Develop segment tags for tos

mauretto78:


Ostico (approved):

lib/Model/LQA/EntryDao.php
```diff
@@ -316,4 +317,55 @@ public function updateRebutted( $id, $isToRebut ) {
...
+        $sql     = "SELECT
+                s.internal_id as content_id,
+                s.id as segment_id,
+                e.severity as severity_label,
+                penalty_points,
+                severities,
+                options as cat_options,
+                label as cat_label
+            FROM
+                matecat.qa_entries e
```
Ostico, line 342: @mauretto78  This schema name is wrong, schema addressing must be removed.

Ostico (changes requested):

lib/Model/LQA/EntryDao.php
mauretto78: @Ostico fixed.
(the brach `mauretto78:develop-segment-tags-for-tos` is actually merged with develop)

Ostico (approved):

PR #2189, opened 2021-09-06
Fix download empty data

mauretto78:


mauretto78:
Composer updated to:
- xliff parser v1.0.81
- subfiltering v1.0.13

PR #2213, opened 2021-10-07
Improving API performance

mauretto78:


Ostico (approved):

Ostico (changes requested):
Cache on getProjectStatsVolumeAnalysis should be invalidated on split/merge

PR #2221, opened 2021-10-14
Fix on APIs

mauretto78:


lib/Controller/API/V2/ProjectCreationStatusController.php
```diff
@@ -23,6 +23,17 @@ class ProjectCreationStatusController extends KleinController {
 
     public function get(){
 
+        // validate id_project
+        if(!is_numeric($this->request->id_project)){
+
+            $response = [];
+            $response[] = new Exception( "ID project is not a valid integer", -1 );
+
+            $this->response->code( 500 );
```
Ostico, line 32: %00 is a wrong HTTP code for this, if this is a "user" request, the validation must return a 400 error code, 

> The HyperText Transfer Protocol (HTTP) 400 Bad Request response status code indicates that the server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).

Ostico (changes requested):

Ostico (approved):

Issue #2270, opened 2021-12-07
fatal_errors.log show "error":"SQLSTATE[42S02]: Base table or view not found: 1146 Table 'matecat.segment_metadata' doesn't exist"


ismail197:
fatal_errors.log show
"error":"SQLSTATE[42S02]: Base table or view not found: 1146 Table 'matecat.segment_metadata' doesn't exist"

PR #2283, opened 2021-12-16
Fix SQL file schema

Ostico:


Ostico:
Ref:
https://github.com/matecat/MateCat/issues/2270

PR #2440, opened 2022-05-24
MMT code update

riccio82:


lib/Controller/AbstractControllers/viewController.php
```diff
@@ -178,6 +179,18 @@ private function setTemplateFinalVars() {
 
     }
 
+    /**
```
Ostico, line 182: @mauretto78 why this piece of code here? should be in the plugin

Ostico (changes requested):

lib/Model/Jobs/JobStruct.php
```diff
@@ -320,6 +335,18 @@ public function isSecondPassReview(){
         return $this->is_review  && $this->_sourcePage == 3;
     }
 
+    /**
+     * @return bool
+     * @throws Exception
+     */
+    public function isArchiveable() {
```
Ostico, line 342: @mauretto78 this should be removed?

Ostico (changes requested):

Ostico (dismissed):

mauretto78 (approved):

Ostico (approved):

PR #2455, opened 2022-06-09
Header fixes

riccio82:


public/js/cat_source/es6/components/segments/Editarea.js
```diff
@@ -413,9 +413,11 @@ class Editarea extends React.Component {
 
     if (changedDecorator) {
       const decorator = new CompositeDecorator(this.decoratorsStructure)
-      this.setState({
-        editorState: EditorState.set(editorState, {decorator}),
-        activeDecorators,
+      setTimeout(() => {
```
riccio82, line 416: @piedicianni This fix the issue in the search feature.

PR #2605, opened 2022-12-28
Fix long query segments

Ostico:


public/js/cat_source/es6/actions/CatToolActions.js
```diff
@@ -274,7 +274,7 @@ let CatToolActions = {
           entries: domains,
         })
       }
-      if (haveKeysGlossary) {
+      if (haveKeysGlossary !== undefined) {
```
Ostico, line 277: @piedicianni @riccio82 please check if this is correct

piedicianni: yes Dom, it's correct!

mauretto78 (dismissed):

public/js/cat_source/es6/actions/CatToolActions.js
Ostico: :+1:

lib/Model/Segments/SegmentDao.php
```diff
@@ -1113,7 +1118,16 @@ public static function getSegmentsForAnalysisFromIdJobAndPassword($idJob, $passw
...
             LEFT JOIN
-                files_parts fp ON fp.id = s.id_file_part        
+                files_parts fp ON fp.id = s.id_file_part     
+            LEFT JOIN (
+                SELECT id_job, id_segment, group_concat( distinct source_page )  as _page
+                FROM segment_translation_events stex
+                JOIN
+                    jobs j ON stex.id_job = j.id
+                WHERE stex.id_job = j.id
+                    AND id_project = :id_project
```
Ostico, line 1128: @mauretto78 this will fail, there is not id_project in thid method signature

piedicianni (approved):

PR #2946, opened 2023-10-13
Fix 100% pre-confirmed segments state

mauretto78:


riccio82:
https://app.asana.com/0/1134617950425092/1204463302993356/f

PR #3431, opened 2024-07-01
Fix TM key rename template

piedicianni:


piedicianni:
Asana: https://app.asana.com/0/1134617950425092/1207684817435395

PR #3441, opened 2024-07-08
Fix tab open with shortcut open settings panel

piedicianni:


piedicianni:
Asana: https://app.asana.com/0/1134617950425092/1207655564092360

PR #3561, opened 2024-09-06
Fix open settings panel - set active incorrect segment footer tab

piedicianni:


piedicianni:
Asana: https://app.asana.com/0/1134617950425092/1208112084155545

PR #3573, opened 2024-09-12
Filters config tab

piedicianni:

PR #3588, opened 2024-09-19
Project template new params

mauretto78:
**ASANA**
https://app.asana.com/0/1134617950425092/1208135944820387
https://app.asana.com/0/1134617950425092/1208135944820383

**NEW TEMPLATE PROJECT PROPERTIES**

The three new template project properties are summarized in the table here:

| parameter name    | type    | required | default value | validation                     | saved in/as |
|-------------------|---------|----------|---------------|--------------------------------|------------------|
| `subject`           | string  | no       | null          |                                | project subject |
| `source_language`   | string  | no       | null          | is a valid language            | job source language |
| `target_language`   | array   | no       | []            | is an array of valid languages | job target language(s) |

**USER METADATA**

To save or update a user metadata call this new endpoint:

```PUT /api/v2/user/metadata```

With the following JSON payload:

```json
{
     "key": "your-key",
     "value": "the-value"
}
```

A response like this will be returned:

```json
{
     "id": 12,
     "uid": 134,
     "key": "your-key",
     "value": "the-value"
}
```

If the `key` already exists, the metadata will be updated.


piedicianni:
FE dependency: https://github.com/matecat/MateCat/pull/3573

PR #3675, opened 2024-11-12
Fix tag size when "Show whitespace characters" is active

piedicianni:
---
  - https://app.asana.com/0/0/1208491821729823


piedicianni:
Asana: https://app.asana.com/0/1200472765951916/1208491821729823

riccio82 (approved):

PR #3903, opened 2025-04-16
`stripTagsPreservingHrefs` improvement 

mauretto78:
Added support for `img` tags

---
  - https://app.asana.com/0/0/1209782451908793


tests/unit/Utils/StripTagsPreservingHrefsTest.php
```diff
@@ -34,4 +34,12 @@ public function testCanStripTagsFromJson() {
 
         $this->assertEquals( $expected, $stripTags );
     }
+
+    public function testCanPreserveImgSrc() {
```
Ostico, line 38: @mauretto78 Can you also add a little test for the href link extraction?

Ostico (changes requested):

Ostico (dismissed):

PR #3907, opened 2025-04-17
Update api_v3_routes.php

Ostico:
Unified controller to split jobs (do not works before merging the qe-workflow-dev branch)


riccio82:
https://github.com/matecat/MateCat/pull/3916

PR #4059, opened 2025-08-08
Correct message for upload file error

mauretto78:
---
  - https://app.asana.com/0/0/1207834449838526

### DIFF
--- a/lib/Model/Conversion/ConversionHandler.php
+++ b/lib/Model/Conversion/ConversionHandler.php
@@ -55,8 +55,6 @@ class ConversionHandler {
     }
 
     public function getLocalFilePath(): string {
-        $this->file_name = html_entity_decode( $this->file_name, ENT_QUOTES );
-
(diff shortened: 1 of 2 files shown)


lib/Model/Conversion/ConversionHandler.php
```diff
@@ -77,6 +77,14 @@ public function processConversion(): void {
         $this->result->setFileName( ZipArchiveHandler::getFileName( AbstractFilesStorage::basename_fix( $this->file_name ) ), $isZipContent );
 
         if ( !file_exists( $file_path ) ) {
+
+            if($_POST['file_name'] !== $this->file_name){
```
Ostico, line 81: Avoid using superglobals inside the low level classes. There are controllers and upload handlers to check for the file name.

mauretto78: @Ostico done, I removed the superglobal $_POST variable.

I moved the file name validation inside `setFileName` method.

PR #4552, opened 2026-04-30
🐛  fix: bypass DAO cache in isSegmentDisabled to prevent stale disabled …

mauretto78:
## Summary

Fix `isSegmentDisabled` returning stale `false` values due to DAO-level cache (1-week TTL) not being invalidated when a segment is disabled.

## Type

- [x] `fix` — bug fix

## Changes

| File | Change |
|------|--------|
| `lib/Controller/Traits/SegmentDisabledTrait.php` | Pass `ttl: 0` to `SegmentMetadataDao::get()` to bypass the DAO cache, which was serving stale results for up to 1 week |

## Testing

Verified that after disabling a segment via `setTranslationDisabled`, the `segment-analysis` API correctly returns `disabled: true` without waiting for cache expiry.

## AI Disclosure

- [x] AI tools were used — details below

GitHub Copilot

## Notes

**Root cause:** `SegmentMetadataDao::get()` defaults to a 604800s (1 week) TTL. When `isSegmentDisabled` queried a segment *before* it was disabled, the DAO cache stored an empty result for 1 week. Subsequent calls to `setTranslationDisabled` wrote the DB row but never invalidated this specific DAO cache entry, so `isSegmentDisabled` kept returning `false`.

**Fix:** Since `isSegmentDisabled` already manages its own Redis cache layer (1-hour TTL) with proper invalidation via `destroySegmentDisabledCache` and `saveSegmentDisabledInCache`, the DAO-level cache is redundant. Setting `ttl: 0` bypasses it.

### DIFF
--- a/lib/Controller/Traits/SegmentDisabledTrait.php
+++ b/lib/Controller/Traits/SegmentDisabledTrait.php
@@ -50,7 +50,7 @@ trait SegmentDisabledTrait
 
         // retrieve from cache fails, we need to check the database and update the cache accordingly
         if (empty($cachedValue)) {
-            $metadataList = SegmentMetadataDao::get($id_segment, 'translation_disabled');
+            $metadataList = SegmentMetadataDao::get($id_segment, 'translation_disabled', 0);
             $metadata = $metadataList[0] ?? null;
             $isDisabled = (($metadata->meta_value ?? '0') === '1');
             $this->_setInCacheMap($cache['key'], $cache['query'], [$isDisabled ? 1 : 0]);


Ostico (changes requested):
## SegmentDisabledTrait — Code Review Analysis

##### Fix for https://github.com/matecat/MateCat/pull/4552

Requested a feature review

##### Prompt Request

> Let's analyze this class: `./lib/Controller/Traits/SegmentDisabledTrait.php`
> 
> Take as reference this one: `./lib/Controller/Abstracts/Authentication/SessionTokenStoreHandler.php`
> 
> The analyzed class is used here:
> - ./lib/Controller/API/App/SetTranslationController.php
> - ./lib/Controller/API/V3/SegmentAnalysisController.php
> - ./lib/Controller/API/V3/CancelRequestController.php
>
> Evaluate the trait for necessity and value, perform a code review, confirm that its style matches the reference implementation, flag any code smells or defects, and propose improvements, simplifications, or removal. Base your recommendations on KISS, YAGNI, and DRY.

---

### What It Does

A trait mixed into controllers that provides a **custom Redis cache layer** for a single boolean question: "is segment X disabled for job Y?" It wraps `SegmentMetadataDao::get($id_segment, 'translation_disabled')` with a hand-rolled read-through cache using `DaoCacheTrait` primitives.

### Verdict: The trait has real value, but its implementation is not totally correct.

The **concept** is valid — 3 controllers need this check, and `CancelRequestController` needs write+invalidation too. However, the **implementation** has some architectural concerns when measured against the reference and KISS/YAGNI/DRY principles.

---

### Concern 1: Redundant Cache Layer (YAGNI + DRY)

This is the core issue. `SegmentMetadataDao::get()` **already caches** via `_fetchObjectMap()` with a 604800s (1 week) TTL:

```php
// SegmentMetadataDao::get() — line 82
$results = $thisDao->setCacheTTL($ttl)->_fetchObjectMap(
    $stmt, SegmentMetadataStruct::class, [$id_segment, $key]
);
```

The trait adds a **second** Redis cache on top with a 3600s (1 hour) TTL. So a read path hits:

1. **Trait's custom cache** (hash map `segment_is_disabled_{job}_{segment}`, 1h TTL)
2. **DAO's built-in cache** (hash map keyed by query+class, 1 week TTL)
3. **MySQL** (only on double cache miss)

The trait's cache is **more restrictive** (1h vs 1 week), which increases DB fallthrough compared to relying on the DAO cache alone — likely the opposite of the intent. The trait's cache is redundant and adds complexity without a measurable benefit, which goes against the YAGNI principle (the extra cache layer may not be needed) and DRY (the caching logic is already provided by the DAO).

### Concern 2: DaoCacheTrait in a Controller Trait (Layering Mismatch)

Compare with the reference:

| Aspect | SessionTokenStoreHandler (reference) | SegmentDisabledTrait |
|--------|--------------------------------------|---------------------|
| Type | **Dedicated class** | Trait mixed into controllers |
| Layer | Service/infrastructure | Controller layer |
| DaoCacheTrait usage | Well-suited — it IS a data store | Unusual — controllers aren't data stores |
| TTL/xFetch config | Set once in constructor | Re-set on every method call via `cacheInit()` |
| Public API | Clean: `set`, `is`, `remove` | Exposed as `protected` methods on controllers |

`SessionTokenStoreHandler` uses `DaoCacheTrait` because it **is** a cache-backed store — it stores login tokens in Redis with no DB backing. That's the intended abstraction.

`SegmentDisabledTrait` uses `DaoCacheTrait` to cache data that **already lives in a DAO with its own cache**. It pulls DAO-level infrastructure (`_setInCacheMap`, `_getFromCacheMap`, `_cacheSetConnection`) into the controller layer. Every controller that uses this trait inherits all of `DaoCacheTrait`'s protected methods and the static `$cache_con` property — which is more surface area than controller classes typically need.

### Concern 3: `cacheInit()` Called on Every Method (vs. Reference Pattern)

```php
// Trait pattern — called in EVERY method
private function cacheInit(): void {
    $this->xFetchEnabled = false;
    $this->setCacheTTL(self::CACHE_TTL);
    $this->_cacheSetConnection();
}
```

The reference sets this **once** in the constructor:

```php
// Reference — set once
public function __construct() {
    $this->cacheTTL = 60 * 60 * 24 * 7;
    $this->xFetchEnabled = false;
}
```

The trait can't use a constructor (it's a trait), so it compensates with `cacheInit()`. This is a sign that a dedicated class would be a better fit here, since a class would handle initialization naturally.

### Concern 4: `destroySegmentDisabledCache` Mixes Concerns

```php
protected function destroySegmentDisabledCache(int $id_job, int $id_segment): void {
    SegmentMetadataDao::delete($id_segment, 'translation_disabled');     // DB write
    SegmentMetadataDao::destroyGetCache($id_segment, 'translation_disabled'); // DAO cache bust
    SegmentMetadataDao::destroyGetAllCache($id_segment);                // DAO cache bust
    // ... plus custom cache bust
}
```

Having a controller trait perform **DB deletes** and **DAO cache invalidation** combines concerns that would be more naturally housed in the DAO or a service layer. Controllers ideally delegate these responsibilities rather than carrying them directly.

### Concern 5: Loose Comparison on Cache Value

```php
return $cachedValue == [1];  // loose comparison
```

`[0] == [1]` → false, `[1] == [1]` → true — it works in practice, but using strict comparison (`=== [1]`) would be safer and clearer about intent.

### Concern 6: The `CACHE_TTL` Constant in a Trait

```php
const int CACHE_TTL = 3600;
```

If another trait mixed into the same controller were to define `CACHE_TTL`, PHP would throw a fatal error. The constant is trait-scoped but has a fairly generic name, which is worth keeping in mind as the codebase grows.

---

### What I'd Recommend

**Option A (cleanest, matches reference pattern): Extract to a service class**

Following `SessionTokenStoreHandler` as the template, create `SegmentDisabledService`:

```php
class SegmentDisabledService {
    // Direct DAO calls, no redundant cache layer
    public function isDisabled(int $id_segment): bool {
        $metadata = SegmentMetadataDao::get($id_segment, 'translation_disabled');
        return $metadata !== null && $metadata->meta_value === '1';
    }
    
    public function disable(int $id_segment): void {
        SegmentMetadataDao::setTranslationDisabled($id_segment);
    }
    
    public function enable(int $id_segment): void {
        SegmentMetadataDao::delete($id_segment, 'translation_disabled');
        SegmentMetadataDao::destroyGetCache($id_segment, 'translation_disabled');
        SegmentMetadataDao::destroyGetAllCache($id_segment);
    }
}
```

This removes: the redundant cache layer, `DaoCacheTrait` from controllers, the `cacheInit()` pattern, and the `id_job` parameter (which is only used in cache key construction — the DAO doesn't need it).

**Option B (minimal change): Keep the trait but strip the custom cache**

If you want to avoid a larger refactor, at minimum removing the `use DaoCacheTrait`, `cacheInit()`, `cacheKeyAndQuery()`, `saveSegmentDisabledInCache()`, and the custom cache bust from `destroySegmentDisabledCache()` would be a meaningful improvement — just call the DAO directly. The DAO's 1-week TTL cache already handles performance.

---

### Summary

| Criterion | Assessment |
|-----------|-----------|
| **Is it needed?** | The **functionality** is needed (3 controllers use it). A **service class** would express it more cleanly than a trait. |
| **KISS** | ❌ Adds a custom cache layer over an already-cached DAO |
| **YAGNI** | ❌ The extra cache layer solves no measurable problem |
| **DRY** | ❌ Duplicates caching logic that `SegmentMetadataDao` already provides |
| **Reference alignment** | ❌ Reference uses a dedicated class; a trait mixing DAO infrastructure into controllers is an unusual pattern |
| **Bugs** | No correctness bugs; the loose comparison and constant naming are minor points to watch |
| **Recommendation** | Extract to a service class (Option A), or at minimum strip the redundant cache (Option B) |

PR #4675, opened 2026-07-09
✨ feat(sse): notify browser tabs live when a segment is disabled or enabled

mauretto78:
## Summary

`/segment/disable` and `/segment/enable` can be called by external company applications, not only through the MateCat browser UI. When that happened while a translator had the project open, the segment's disabled state never updated in the UI without a manual refresh. This PR emits a live Socket.IO notification so every open tab on the project reflects the change immediately.

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Changes

| File | Change |
|------|--------|
| `lib/Controller/API/V3/CancelRequestController.php` | `performChecks()` now returns the fetched `JobStruct`; publish a `segment_disabled`/`segment_enabled` AMQ message (routed by `id_project`) only when the segment's disabled state actually transitions |
| `nodejs/src/amq/MessageHandler.js` | New `segment_disabled`/`segment_enabled` cases routing by `data.id_project` |
| `nodejs/src/Application.js` | Accept an `x-projectid` auth header and join a matching Socket.IO room, mirroring the existing `x-jobid`/jobId handling |
| `public/js/hooks/useSocketLayer.js` | Send `x-projectid` header on connect |
| `public/js/sse/SocketListener.js` | Thread `projectId` into the socket connection; add `segment_disabled`/`segment_enabled` event handlers that live-patch segment state and show a toast |
| `public/js/constants/SegmentConstants.js`, `public/js/actions/segmentDispatchActions.js`, `public/js/actions/SegmentActions.js` | New `SET_SEGMENT_DISABLED` Flux action, mirroring the existing `SET_SEGMENT_STATUS` pattern |
| `public/js/stores/SegmentStore.js` | New `setDisabledMetadata()` reducer that patches a segment's `translation_disabled` metadata in place |
| `tests/unit/Core/Controllers/CancelRequestControllerTest.php` | New tests asserting the publish payload/routing on state transitions and no-ops; hand-written `Stomp\Client`/`Stomp\StatefulStomp` fakes to avoid a PHP GC/destructor hazard with PHPUnit mocks |
| `nodejs/tests/Application.test.js`, `nodejs/tests/MessageHandler.test.js` | Coverage for the new `x-projectid` header/room-join and `id_project` message routing |
| `public/js/actions/SegmentActions.test.js` | Coverage for the new `updateSegmentDisabledState` dispatch |

## Testing

`vendor/bin/phpunit tests/unit/Core/Controllers/CancelRequestControllerTest.php` — 36/36 passing. Full-repo `vendor/bin/phpstan analyse --configuration=phpstan.neon` — 0 errors. Node.js `npx jest` — 73/73 passing across all suites. Frontend `yarn test` could not be run in the dev sandbox used to prepare this PR (pre-existing Jest/Babel toolchain issue unrelated to this change, reproducible on an untouched test file) — please confirm `yarn test` passes in CI.

## AI Disclosure

Claude Code (claude-sonnet-5)

## Notes

Design decision: routing by `id_project` (rather than the existing job-scoped room) was a deliberate choice — `translation_disabled` metadata is stored per-segment, not per-job, so a disable/enable notification is inherently project-wide, not job-scoped.

---
  - https://app.asana.com/0/0/1214028216264288


lib/Controller/API/V3/CancelRequestController.php
```diff
@@ -120,7 +134,7 @@ public function cancelRequest(): void
      * @throws NotFoundException If the job or segment is not found.
      * @throws Exception If the user is not the owner or part of the team, or if the segment status is not "new".
      */
-    private function performChecks(int $id_job, string $password, int $id_segment, string $route): void
+    private function performChecks(int $id_job, string $password, int $id_segment, string $route): ?JobStruct
```
Ostico: Why this return type needs to be nullable?
There is a null check already in place:

```
        if (null === $job) {
            throw new NotFoundException('Job not found.');
        }
````

Returning nullable force to write the assertion in line 79.

Please check:
https://github.com/matecat/MateCat/pull/4675/changes#diff-674dde540e8fcb6d3c4842aca4295bd6f095888a5badeb19ea5166367413ac5cR147

Ostico (changes requested):

riccio82 (dismissed):

lib/Controller/API/V3/CancelRequestController.php
mauretto78: Good catch — fixed in 7ed23d9. Moved rate-limit handling out of `performChecks()` into a new `isRateLimited()` helper called from `enableRequest()`/`cancelRequest()` (mirroring the pattern in `KeyCheckController`), so `performChecks()` now always returns a non-null `JobStruct` or throws. This also removes the `assert($job !== null)` at each call site and resolves the CyclomaticComplexity/NPath findings from GitHub Advanced Security on the same method. Existing 36 tests pass unchanged.

Ostico (approved):

PR #4710, opened 2026-07-24
🐛 fix(mymemory): hydrate source and target language codes on TM matches

mauretto78:
## Summary

MyMemory TM matches were losing their source/target language codes, and `segment`/
`translation` could end up `null` when the API response omitted those keys. This fixes
`MyMemory::_decode()` and `GetMemoryResponse::buildMyMemoryMatch()` to hydrate `source`/
`target` from the engine config and to always default `segment`/`translation` to an empty
string.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Changes

| File | Change |
|------|--------|
| `lib/Utils/Engines/MyMemory.php` | inject `source`/`target` from `$this->_config` into decoded matches; default `segment`/`translation` to `''` when absent |
| `lib/Utils/Engines/Results/MyMemory/GetMemoryResponse.php` | hydrate `source`/`target` (and null-safe `tm_properties`) onto `Matches` in `buildMyMemoryMatch()` |
| `tests/unit/Core/Engines/Results/MyMemory/GetMemoryResponseTest.php` | new regression tests for missing `segment`/`translation`, missing `tm_properties`, and `source`/`target` hydration |

## Testing

Ran the full suite in Docker: 9222 tests, 30003 assertions. 4 pre-existing failures in
`CommentControllerTest` (broker-unavailable scenarios) are environment-dependent and
unrelated to this change — they also fail on `develop` in this environment since the
ActiveMQ broker container is reachable rather than down.

`phpstan` run clean (0 errors) on all three changed files.

## AI Disclosure

Claude Code (claude-sonnet-5)

## Notes

None.

---
  - https://app.asana.com/0/0/1216294000966428

### DIFF
--- a/lib/Utils/Engines/MyMemory.php
+++ b/lib/Utils/Engines/MyMemory.php
@@ -176,9 +176,11 @@ class MyMemory extends AbstractEngine
 
                 if (!empty($decoded['matches'])) {
                     foreach ($decoded['matches'] as $pos => $match) {
-                        $decoded['matches'][$pos]['segment'] = $match['segment'];
-                        $decoded['matches'][$pos]['translation'] = $match['translation'];
+                        $decoded['matches'][$pos]['segment']     = $match['segment'] ?? '';
+                        $decoded['matches'][$pos]['translation'] = $match['translation'] ?? '';
                         $decoded['matches'][$pos]['target_note'] = $match['target_note'] ?? '';
+                        $decoded['matches'][$pos]['source']      = $this->_config['source'] ?? $match['source'] ?? '';
+                        $decoded['matches'][$pos]['target']      = $this->_config['target'] ?? $match['target'] ?? '';
                     }
                 }
 
--- a/lib/Utils/Engines/MyMemory.php
+++ b/lib/Utils/Engines/MyMemory.php
@@ -179,8 +179,8 @@ class MyMemory extends AbstractEngine
                         $decoded['matches'][$pos]['segment']     = $match['segment'] ?? '';
                         $decoded['matches'][$pos]['translation'] = $match['translation'] ?? '';
                         $decoded['matches'][$pos]['target_note'] = $match['target_note'] ?? '';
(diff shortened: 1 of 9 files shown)


lib/Utils/Engines/MyMemory.php
```diff
@@ -176,9 +176,11 @@ protected function _decode(mixed $rawValue, array $parameters = [], ?string $fun
 
                 if (!empty($decoded['matches'])) {
                     foreach ($decoded['matches'] as $pos => $match) {
-                        $decoded['matches'][$pos]['segment'] = $match['segment'];
-                        $decoded['matches'][$pos]['translation'] = $match['translation'];
+                        $decoded['matches'][$pos]['segment']     = $match['segment'] ?? '';
+                        $decoded['matches'][$pos]['translation'] = $match['translation'] ?? '';
                         $decoded['matches'][$pos]['target_note'] = $match['target_note'] ?? '';
+                        $decoded['matches'][$pos]['source']      = $this->_config['source'] ?? $match['source'] ?? '';
```
Ostico: Here, we are using the match language returned by MyMemory as a fallback—prioritizing it over the language defined in the class configuration, which might eventually be populated with Matecat's source/target values. Currently, this fallback has no effect because `$this->_config['source'/'target']` are always null, but if we start using those values ​​later, the feature will break. The two parameters need to be swapped.

Ostico (changes requested):

lib/Utils/Engines/MyMemory.php
mauretto78: Good catch, fixed — swapped the precedence in `MyMemory::_decode()` so `$match['source']`/`$match['target']` now win over `$this->_config`, with config only used as a fallback when the match omits them. Added two regression tests in `DecodeMyMemoryTest` covering both directions (match wins when present, config used when absent). Pushed as 539447fcf7.

mauretto78:
@Ostico Addressed your review comment — fixed the precedence swap in `MyMemory::_decode()` and added regression tests. Ready for another look.

PR #4712, opened 2026-07-24
🐛 fix(cattool): read job-scoped mandatory issues instead of project snapshot

mauretto78:
## Summary

Selecting a per-job mandatory-issues override (e.g. "only R2") had no effect on segment
approval: the enforcement check and the settings-panel gate condition both read the
project-level metadata snapshot taken once at project creation, instead of the job-level
value that the settings panel actually updates. This fixes both to read the job-scoped
value, and refreshes the cached job metadata immediately after saving so an already-open
editor session picks up the change without a page reload.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Changes

| File | Change |
|------|--------|
| `public/js/actions/SegmentActions.js` | Read `job.mandatory_issues` instead of the stale `project.mandatory_issues` snapshot when gating mandatory-issue enforcement on approval |
| `public/js/pages/CatTool.js` | Fix the same project/job scope mismatch in the gate condition that populates the settings-panel template state |
| `public/js/components/settingsPanel/Contents/EditorOtherTab/EditorOtherTab.js` | Refresh `CatToolStore`'s cached job metadata right after `updateJobMetadata` succeeds, so an open editor session applies the new setting without reload |
| `public/js/actions/SegmentActions.test.js` | Update mocks to job-scoped metadata; add a regression test for a per-job override coexisting with a wider project-level default |

## Testing

Manually verified in the editor: set mandatory issues to "Only R2" on a job, confirmed
approving a modified segment with no issues no longer forces the issues panel open in R1,
while it still does in R2.

## AI Disclosure

Claude Code (claude-sonnet-5)

## Notes

None.

### DIFF
--- a/public/js/actions/SegmentActions.js
+++ b/public/js/actions/SegmentActions.js
@@ -266,8 +266,7 @@ const SegmentActions = {
        If is an ICE we allow to change the translation because is not possible to add an issue
      */
 
-    const mandatoryIssues =
-      CatToolStore.getJobMetadata().project.mandatory_issues
+    const mandatoryIssues = CatToolStore.getJobMetadata().job.mandatory_issues
 
     const currentRevisionKey = `r${config.revisionNumber}`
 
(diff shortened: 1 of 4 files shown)


mauretto78:
Superseded by #4713, opened from matecat/fix-mandatory-issues-job-scope instead of a fork branch.

Issue #47, opened 2016-03-01
Status of segment in Export Job TMX missing


pepc:
Currently there is no way to know if a segment in a exported job TMX is translated or approved.

Sometimes when you QA a translation, it is useful to know the segments status, in order to make a decision when several options are available.

I suggest to add a _status_ metadata to the segment.  For example:
```
<prop type="x-MateCAT-status">translated</prop>
<prop type="x-MateCAT-status">approved</prop>
```

PR #4736, opened 2026-08-01
🔒 security(auth): harden the session and cookie identity path, and finish the $_SESSION conversion

Ostico:
## Summary

Completes the move of `$_SESSION` behind the injected `SessionStore`, and carries two bug fixes found while doing it. The `NoDirectSessionSuperglobalRule` allowlist is now down to the adapter alone, so PHPStan passing is what proves `lib` is clean rather than a hand-maintained list being trusted.

The two fixes are not incidental tidying — both were shipping wrong behaviour:

- **Cookie-authenticated `/api/v2` and `/api/v3` requests answered `401 Invalid Login`** while presenting a valid, unexpired token. Those routes are plain `KleinController`s, so they hold a `StatelessSessionStore` whose `get()` throws by design; `authenticate()` consulted the session cache as the *first* step of its cookie branch, and the surrounding `catch (Throwable)` swallowed the violation into "not authenticated". Api-key callers were unaffected, which is why it went unnoticed.
- **Worker-driven chunk-review updates were attributed to user `0`.** The listeners read the actor from `$_SESSION`, and `BulkSegmentStatusChangeWorker` — the main path into them — has no session.

Since then the branch also carries **a security pass over the authentication and session path**, 13
further commits. That work was not planned when this PR opened; it came out of reviewing the code this
PR moves. Reviewers should treat the two halves separately — the `$_SESSION` conversion is a
restructure, the second half changes security behaviour on every authenticated request.

The security half, one line each:

| Commit | What was wrong |
|--------|----------------|
| `15c0d75e` | The session serialized the whole `UserStruct` — password hash, salt, confirmation token — into Redis for the session's idle lifetime, unpurged on password change. |
| `80c7217b` | `setcookie()` only queues a header, so a login immediately followed by `fromRequest()` authenticated the *previous* occupant. Account switch stayed in the old account. |
| `ed5f01c1` | `GET /login/socket` minted a signed identity JWT from `session['uid']` alone. Revoked users kept receiving socket credentials until the session died of idleness. |
| `cbe2c488` | A login over a live cookie abandoned that token in the ring for its full lifetime. Now chained so the first renewal retires it. |
| `c5aafde3` | Logout cleared the session before revoking. A Redis failure left "signed out" with a live cookie and a live token. |
| `1c8bb061` | `REDIS_MODE=replication` with no `?role=master` reads from a random replica and fails every write. Documentation said position decides; it does not. |
| `2610d28b` | Two docblocks written earlier in this stack were factually wrong. |
| `42f2c0b8` | `PhpSessionStore::regenerateId()` had no callers; the helper carried its own copy of the guards. |
| `92b8aa13` | `AuthCookie` called `session_destroy()` from inside its cookie-clearing helper, so a cookie that merely failed to parse ended the session. |
| `ebaf8102` | **Revocation was not final.** A renewal writing after a `DEL` recreated the ring with a token minted *after* the user was revoked. |
| `3b692bd0` | `getValidatedInstanceFromString()` skipped validation entirely when the secret was empty — signature *and* expiry. |
| `ba1d9fec` | Team invites tested the team's pending set for emptiness, so withdrawing one invitation did nothing while any other stood. |
| `eddeed31` | A different user taking over a session inherited its contents — cart, GDrive tokens, redeemable project. |
| `f1d92ddf` | An identity switch this process did not perform was accepted, so a planted cookie silently moved a user into somebody else's account. |

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

`fix` rather than `refactor`: the bulk is a restructure, but two of the commits change observable behaviour, so labelling this "no behavior change" would misdirect review.

## Changes

| File | Change |
|------|--------|
| `AuthenticationHelper::cachedSessionUser()` | Guard the session-cache read with `sessionIsActive()`, the same condition that already gates the write in `setUserSession()`. No session is a cache miss, not an error. |
| `ChunkReviewModel`, `IChunkReviewModel` | Acting `UserStruct` becomes a required parameter on the five dispatching methods — not the constructor, since the same class serves read-only callers (`QualitySummary`, `QualityReportModel`) that never dispatch. |
| `ChunkReviewUpdatedEvent`, `SetTranslationCommittedEvent`, `PostJobSplittedEvent`, `PostJobMergedEvent` | Carry the actor as a typed field, so listeners stop reaching for request-global state. |
| `BatchReviewProcessor` | Takes the actor; drops a dead `?? $translationEvent->getUser()` fallback (every event in a batch is built with the same user). |
| `BulkSegmentStatusChangeWorker` | Throws `EndQueueException` when the enqueuing user no longer exists, instead of proceeding unattributed. |
| `TranslationVersionsHandler` | Throws when `$params['user']` is not a `UserStruct`; both production callers already pass `$this->user`. |
| `JobSplitMergeManager`, `JobSplitMergeService`, `SplitJobController` | Thread a separate actor for split/merge. Deliberately **not** `SplitMergeProjectData::$uid`, which also gates translator re-invitation. |
| `CreateProjectController` | Last direct `$_SESSION` read in `lib` moves to the store, behind a `gdriveSessionData()` seam. |
| `phpstan.neon` | All 18 `lib/` allowlist entries removed, plus `plugins/aligner` as dead config. Only the adapter remains. |
| `InvitedUserTest` | Two cases repaired: they seeded `$_SESSION` while handing the subject an empty `ArraySessionStore`, so they short-circuited and never tested their named condition. |
| `internal_scripts` gitlink | `revision:recount` gains a required `uid` argument. |
| `plugins/translated` gitlink | Plugin reads the actor from the event instead of the session. |

Security half:

| File | Change |
|------|--------|
| `AuthenticationHelper::setUserSession()` | Clears the session when a *different* uid takes it over, and refuses a switch this process did not perform. Rotation moved to the store. |
| `AuthenticationHelper::destroyAuthentication()` | Revokes before ending the session, and ends the whole session rather than clearing the `uid` marker. |
| `AuthCookie` | `$_COOKIE` kept in step inside the emitter; `prev` chaining at login; renewal re-checks the ring after minting; records that it issued credentials; no longer touches the PHP session. |
| `SimpleJWT::getValidatedInstanceFromString()` | Throws on an empty secret instead of skipping signature and expiry checks. |
| `RedisHandler` | Replication mode refuses a server set that names no master; the Mode 3 documentation is corrected. |
| `PendingInvitations` | `isPending()` (SISMEMBER on this team and address) replaces an emptiness test; `hasPendingInvitation()` renamed `listPendingInvitations()` at its three display call sites. |
| `SessionStore` + 3 implementers | New `clear()` — empty the contents, keep the storage usable. Distinct from `destroy()`, which ends the session so the following write neither rotates nor persists. |

## Testing

```
OK (9576 tests, 30728 assertions)
[OK] No errors
```

Note on the PHPStan line: this repository has **no** `phpstan-baseline.neon`. The 0 errors are at level 8 with no baseline at all, which is stricter than the checkbox wording.

Each new test was verified to fail before its fix, and two claims were confirmed by mutation rather than argument:

- Reverting `gdriveSessionData()` to the superglobal fails the seam test with `Failed asserting that null is identical to Array ('access_token' => 'tok', ...)`.
- In its pre-repair shape, `hasPendingInvitationsReturnsFalseWhenNoMembers` passes **green** with a pending member present — proof it could not fail for its own subject.

**Manual testing — the fix confirmed over HTTP.** The 401 was originally reproduced against a running instance with a browser login cookie on `GET /api/v2/teams/12/members`, and the deployed code was confirmed to be this branch by matching stack-trace line numbers. After the fix, the identical request — same cookie, same endpoint, same instance — returns `200`:

```
GET /api/v2/teams/12/members     before: 401 Invalid Login     after: 200
{"members":[{"id":13,"id_team":12,"user":{"uid":36, ... }, "projects":1184}],
 "pending_invitations":[]}
```

Identity resolved to uid 36 from the login cookie alone, on a stateless controller with no session available — which is precisely the path that was raising `StatelessSessionViolation` into the swallowing `catch`.

This had to be a cookie request. An api-key call proves nothing here: `authenticate()` returns in the `validKeys` branch before the session is touched, so the api-key path never exercised the bug and never exercises the fix.

The security half adds 26 tests (9550 → 9576) and **every production change in it is
mutation-checked**: the line is reverted, and a named test must fail. Where a rule has two independent
halves, each half is killed by its own case — for example dropping the session-clear fails the
takeover test while dropping its null guard fails the anonymous-transition test, which is what proves
the guard is load-bearing rather than defensive.

Whole-repo PHPStan was re-run at the end, not only per file, because `SessionStore` gained an
interface method — an unimplemented one would be a fatal error rather than a test failure.

Two live checks on the running instance: the token ring holds only `uid` plus OAuth XSRF tokens with
zero credential matches across the session database, and every authenticated route answers 200 after
login with no 401 or 500.

## AI Disclosure

Claude Code (claude-opus-5)

## Notes

**Stacked PR.** Base is `refactor/session-store`, not `develop`.

**Two submodule PRs must merge together with this one.** This branch bumps both gitlinks, so merging it alone leaves them pointing at commits that are not on either submodule's default branch:

- `matecat/translated_plugin` #66
- `matecat/internal_scripts` #47

**`revision:recount` is a breaking invocation change.** Any runbook or cron entry calling it without a `uid` now fails with a missing-argument error rather than silently attributing the repair to user `0`. That failure is the intent.

**Verified over HTTP** — see Testing. The endpoint that returned `401` with a valid cookie now returns `200`, so this is confirmed end to end and not only by unit test.

**Unrelated but worth knowing:** the local dev instance's TLS certificate has expired, so any client that validates certificates cannot reach it without skipping verification. Not caused by this PR, but it will affect anyone trying to reproduce the check above.

**The aligner is not a blocker, contrary to earlier notes.** Its allowlist entry was dead on two counts: `plugins/aligner` is in `excludePaths` so the rule never analysed it, and the file no longer names `$_SESSION` at all. The only occurrence left under `plugins/aligner/lib` is prose in a `ConfigsController` docblock. No companion aligner PR is needed.

**Known coverage gap.** The kafka payload block in the plugin's `chunkReviewUpdated()` is uncovered — reaching it needs the static `FeatureSet::forProject` with a database, a `review_extended` project, an `@translated.net` customer and a DB-backed `QualitySummary::render()`. Pre-existing; the two changed lines were uncovered before as well.

**That follow-up landed here.** An earlier revision of these notes listed the session serializing the
whole `UserStruct` as out of scope. `15c0d75e` does it: the session now holds `uid` and nothing else,
and `cachedSessionUser()` is gone with it.

**Two deliberate deviations from the plan, both recorded in the commits.** The revoke/renew race is
closed with a post-write re-check plus compensation rather than an atomic Lua `EVAL`: Lua would have to
reproduce the exact stored encoding `DaoCacheTrait` writes, and getting that wrong logs out every live
session on deploy. It covers every ordering of revoke-versus-renew; the residual hole is a process
death between two adjacent statements. And session destruction moved out of `AuthCookie` instead of
injecting a store into it, which means a cookie that fails to parse no longer ends the session — the
request is anonymous either way, since identity comes from the ring.

**`__Host-` cookie prefix: evaluated and rejected, do not retry as-is.** It requires a host-only
cookie, but `ENABLE_MULTI_DOMAIN_API` means credentialed XHR to `*.ajax.<host>` (`CorsHandler` sends
`Access-Control-Allow-Credentials: true`), so a host-only cookie would answer 401 on every sharded
call. What actually bounds cookie-tossing here is `COOKIE_DOMAIN = "dev.matecat.com"`: it reaches the
shards beneath that domain and is never sent to a sibling such as `blog.matecat.com`. **That is a
property of the config value — widening it to the registrable domain reopens the exposure.** A
prerequisite for the prefix would be moving shard authentication off the cookie.

**Still outstanding.** Dropping `DaoCacheTrait` from `SessionTokenStoreHandler` is deliberately not
here: it is the one change that can log out every live session at deploy, because the stored encoding
`serialize([$cookie])` must be preserved exactly. Retired-token reuse detection is also unbuilt — an
attacker actively using a captured token renews it indefinitely, and the retirement two generations
later lands on the victim's token instead of theirs.


Ostico:
## Test adequacy gate — why `test-guard-exception` is applied

The aggregate diff coverage is **87%**, above the 80% threshold. The gate fails on individual files, and
those fall into two genuinely different groups. Applying the label is not a claim that all of them are
uncoverable — that would be false, and the second group below says so explicitly.

Before applying it I closed the one gap that was plainly mine to close: the new stateful subclass
`lib/Controller/API/App/SplitJobController.php` was at **0%** and is now at **100%** (1/1 statements),
which also moved `V2/SplitJobController.php` and `TeamsInvitationsController.php` into the pass column.

### Structurally impossible — the reason the gate cannot go green

| File | Why |
|---|---|
| `lib/Controller/API/V2/JobMergeController.php` | **Deleted in this PR.** Reported as "not in coverage report" and "No matching test file found". A file that no longer exists cannot be covered, and its test was deleted with it. |
| `lib/Routes/api_v2_routes.php`, `api_v3_routes.php`, `app_routes.php` | Declarative route tables — a flat sequence of `route(...)` registrations executed at include time. No test in this repository targets a route file, and there is no naming convention that could match one. 0% by construction. |
| `lib/Utils/Session/SessionStore.php` | An interface. No executable statements exist to cover. |
| `lib/Utils/Session/StatelessSessionViolation.php` | An exception class with an empty body. |
| `tests/unit/Core/Model/JobSplitMerge/TestableJobSplitMergeService.php` | A **test helper**, flagged for having "no matching test file". A test file requiring its own test file is a category error. |

`AuthenticationHelper.php`, `SignInController.php` and `Cart.php` are also flagged "No matching test file
found" by a filename heuristic. `AuthenticationHelper` is covered by
`AuthenticationHelperRefactoredTest` — the heuristic simply does not match that name.

### Real gaps, deliberately deferred — not covered by the claim above

These are genuinely coverable and are **not** being excused as impossible. They are the session-store
conversion touching code whose existing coverage was already thin, and closing them properly needs
session-backed and DB-backed fixtures rather than another reflection double:

| File | Diff coverage |
|---|---|
| `lib/Controller/API/App/Authentication/LoginController.php` | 78% — 2 points short |
| `lib/Controller/Abstracts/BaseKleinViewController.php` | 62% |
| `lib/Model/JobSplitMerge/JobSplitMergeManager.php` | 50% |
| `lib/Model/JobSplitMerge/JobSplitMergeService.php` | 43% |
| `lib/Controller/API/App/OutsourceToController.php` | 0% |
| `lib/Controller/Views/OutsourceTo/AbstractController.php` | 0% |

The outsource pair is the honest one to call out: those files are where the cart cache lives, this PR is
what makes the previously-inert invalidation actually run, and that path has no coverage. It is recorded
as follow-up rather than fixed here because the correct fix is to move that cache off the session
entirely, which is a separate change.

Local verification for the whole tree: `OK (9545 tests, 30656 assertions)`, whole-repo PHPStan
`[OK] No errors` at level 8 with no baseline.

riccio82 (approved):

riccio82 (approved):

PR #4750, opened 2026-08-05
♻️ refactor(features): remove feature hooks left without implementers

mauretto78:
## Summary

Three feature-hook event classes were dispatched from core but implemented only by the Airbnb plugin,
whose handlers were a no-op, inert, and superseded by a core setting respectively. With the plugin
handlers gone (matecat/airbnb_plugin#122) the events, their dispatch sites and the plumbing that fed
them have no remaining purpose, so they are removed here.

Also drops `BaseFeature::getBuildFiles()`, which no plugin or core code has ever called.

## Type

- [ ] `fix` — bug fix
- [x] `refactor` — restructure without behavior change
- [ ] `test` — test coverage

## Changes

| File | Change |
|------|--------|
| `lib/Model/FeaturesBase/Hook/Event/Filter/ProjectUrlsEvent.php` | deleted — sole implementer was an identity pass-through |
| `lib/Model/FeaturesBase/Hook/Event/Filter/FilterRevisionChangeNotificationListEvent.php` | deleted — sole implementer needed a config key that exists in no environment |
| `lib/Model/FeaturesBase/Hook/Event/Filter/AnalysisBeforeMTGetContributionEvent.php` | deleted — sole implementer duplicated the per-project `enable_mt_analysis` setting |
| `lib/Controller/API/V2/UrlsController.php`, `lib/View/API/V2/Json/Job.php` | drop the `ProjectUrls` dispatch; render the formatted object directly |
| `lib/View/API/V2/Json/Job.php`, `lib/View/API/V3/Json/Chunk.php` | `fillUrls()` no longer needs the `FeatureSet` argument — parameter and call sites updated |
| `lib/Plugins/Features/ReviewExtended/ReviewedWordCountModel.php` | drop the recipient-list dispatch from `_sendNotificationEmail()` |
| `lib/Utils/AsyncTasks/Workers/Analysis/TMAnalysis/Service/EngineService.php` | drop the pre-MT dispatch from `getMTTranslation()` |
| `lib/Plugins/Features/BaseFeature.php` | remove `getBuildFiles()` — no callers anywhere |
| `tests/unit/Core/**` | drop the cases covering the removed events and helper; update the two `fillUrls()` overrides for the new signature |
| `plugins/airbnb` | submodule bumped to the matching plugin commit |

## Testing

PHPStan level 8 over the whole repository → **0 errors**.
The directly affected suites — `UrlsControllerTest`, `BaseFeatureTest`, `FilterEventSubclassTest` —
pass (**91 tests, 180 assertions**), as does `plugins/airbnb/tests` (**89 tests, 108 assertions**).

The full suite on the development machine reports pre-existing
`SQLSTATE[HY000] [2002] Connection refused` errors because no database is reachable from the
container. The only two genuine assertion failures, both in
`EngineServiceUnitTest::getMTTranslation_*`, were verified to fail identically at `HEAD` with these
changes stashed, so they are unrelated.

## AI Disclosure

Claude Code (claude-opus-5)

## Notes

**Merge order.** matecat/airbnb_plugin#122 must merge first. Until it does, the submodule gitlink here
points at a commit that lives on the plugin's `remove-obsolete-code` branch rather than its `master`.

**One intentional behaviour change**, inherited from the plugin side: the Airbnb handler used to flip
the engine's analysis flag back to `false` after `EngineService` had set it, which made the
skip-analysis gate in `AbstractEngine::call()` unreachable and forced MMT to be queried during
analysis regardless of the project setting. Airbnb projects will now honour their own
`enable_mt_analysis` value like every other customer.

**Follow-up, not in this PR.** The plugin's `characterLengthCount()` hook stays for now.
`character_counter_mode` (`google_ads` / `exclude_cjk` / `all_one`) already generalises that
behaviour, but it is read only by `public/js/utils/charsSizeCounterUtil.js` —
`SizeRestriction::getCleanedStringLength()` still hardcodes `mb_strlen + CJK + emoji` and ignores the
job metadata. Teaching the PHP counter to honour the setting would let the last plugin hook and
`CharacterLengthCountEvent` go too.

---
  - https://app.asana.com/0/0/1216847683605440


lib/Model/FeaturesBase/Hook/Event/Filter/FilterRevisionChangeNotificationListEvent.php
Ostico: This event is still used in [Airbnb](https://github.com/matecat/airbnb_plugin/compare/e1fc106744273d7fc6cb0878d2893f7529fbcc49...4012f0ea8ac5942de3ce9599f9c0d0006d952748#diff-837ca71ad44e8dcd4b15e85be477e2e780b0b5995d051f44090ea29702c8e862L95) plugin , removing it remove a real functionality. Re-check

lib/Plugins/Features/ReviewExtended/ReviewedWordCountModel.php
```diff
@@ -386,10 +384,6 @@ private function _sendNotificationEmail(array $finalRevisions, array $chunkRevie
             ];
         }
 
-        $filterRevisionChangeNotificationListEvent = new FilterRevisionChangeNotificationListEvent($emails);
```
Ostico: Here it is used:

https://github.com/matecat/MateCat/pull/4750/changes#r3872133651

lib/Model/FeaturesBase/Hook/Event/Filter/AnalysisBeforeMTGetContributionEvent.php
Ostico: This seems still used in the Airbnb plugin [Airbnb](https://github.com/matecat/airbnb_plugin/compare/e1fc106744273d7fc6cb0878d2893f7529fbcc49...4012f0ea8ac5942de3ce9599f9c0d0006d952748#diff-837ca71ad44e8dcd4b15e85be477e2e780b0b5995d051f44090ea29702c8e862L409) even if MMT is an obsolete engine and it is not practically used anymore.

plugins/airbnb
Ostico: Please check [Airbnb](https://github.com/matecat/airbnb_plugin/compare/e1fc106744273d7fc6cb0878d2893f7529fbcc49...4012f0ea8ac5942de3ce9599f9c0d0006d952748#diff-a268e09062a544fc301c0baf1168add634b0e6228c1faba2dc40f410a0a7e534L24) routes, css and js should now be set from yarn and vite but before removal must be check.

PR #4768, opened 2026-08-18
♻️ refactor(frontend): migrate class components to hooks (waves 1-4d)

riccio82:
## Summary

Converts **47** React class components to function components with hooks — every class component
in `public/js` — and removes the prototype patching that made the last two of them unconvertible.
Deployment-specific behaviour now replaces a member of a plain object core exports, which is the
pattern this codebase already used for segment context and footer tabs, extended to the four points
that lost their prototype home.

Each conversion was independently verified (including break/restore mutation testing against the
existing suite) and reviewed before being committed. A handful of latent, previously-untested bugs
were uncovered during conversion and fixed naturally as part of the rewrite — see Notes.

This PR targets `coverage-improvement-plan` (see Notes on base branch).

## Type

- [ ] `fix` — bug fix
- [x] `refactor` — restructure without behavior change
- [ ] `test` — test coverage

### Waves 1–3 — modals, shared UI, header, language selector, review sidebar (26 files)

| File / group | Change |
|------|--------|
| `common/TextField.js`, `common/WrapperLoader.js` (+test) | Convert to function components |
| `header/cattol/{BulkSelectionBar,QAComponent,search/Search,SubHeaderContainer}` (+tests) | Convert to function components; suppress a stale `no-unused-vars` warning on QAComponent's preserved `allowHTML` |
| `languageSelector/{LanguageSelector,LanguageSelectorList,LanguageSelectorSearch}` (+tests) | Convert to function components; fix a `mousedown` listener leak in `LanguageSelectorSearch`; add regression tests for Enter-during-search and querySearch/callback staleness |
| `modals/{AlertModal,ConfirmMessageModal,CopySourceModal,FatalErrorModal,JobMetadataModal,ModalWindow,RevisionFeedbackModal,ShareTmModal,ShortCutsModal,SuccessModal}` (+tests) | Convert to function components; sanitize `JobMetadataModal` instructions and fix a link-rewrite crash; add regression tests pinning `AlertModal`'s raw-HTML rendering and `ModalWindow`'s listener count on unmount |
| `projects/Header.js` (+test) | Convert to function component |
| `review/TranslationIssuesSideButton.js`, `review_extended/{ReviewExtendedTranslationIssuesSideButton,ReviewExtendedPanel,ReviewExtendedIssuePanel,ReviewExtendedIssuesContainer}` (+tests) | Convert to function components; fix two untested `addListener`/`removeListener` reference-mismatch leaks; add regression tests pinning the listener and context-staleness fixes |

### Wave 4 — the segment editor cluster (18 files)

| File / group | Change |
|------|--------|
| `segments/{SegmentCommentsContainer,SegmentsCommentsIcon,SegmentHeader,SegmentWarnings,SegmentWrapper}` (+tests) | Convert to function components |
| `segments/{SegmentFooterMultiMatches,SegmentFooterTabConcordance,SegmentFooterTabConflicts}` (+tests) | Convert to function components; port `shouldComponentUpdate` to `memo` comparators |
| `segments/GlossaryComponents/{GlossaryHighlight,QaCheckGlossaryHighlight,QaCheckBlacklistHighlight}.component.js`, `segments/LexiqaHighlight/LexiqaHighlight.component.js`, `segments/TooltipInfo/{TooltipInfo,LexiqaTooltipInfo}.component.js`, `segments/utils/DraftMatecatUtils/TagMenu/TagBox.js` (+tests) | Convert the decorator/tooltip leaf components |
| `segments/SegmentPlaceholderLite.js`, `segments/SegmentTarget.js` (+tests) | Convert to function components (Wave 4d) |
| `segments/SegmentSource.js` (+test) | Convert to a function component; route internal cross-method calls through `instanceRef` so `jest.spyOn` on any instance method still observes internally-triggered calls |
| `segments/Editarea.js` (+test) | Convert to a function component — the largest file in the migration (13 state fields, 47 exposed methods, 6-concern `componentDidUpdate`) |
| `segments/TagEntity/TagEntity.component.js` (+2 tests) | Convert to a function component (Wave 4e); rewrite the suite as behaviour tests first, then fix two update bugs it exposed — see Notes |

### Retiring prototype patching (new)

The last two class components could not be converted while deployment plugins reached into them by
assigning to `SomeComponent.prototype.someMethod`. A function component has no prototype, so that
coupling had to go. It is replaced with no mechanism at all: core calls a member of a plain
exported object, and a deployment assigns over that member — exactly what `globalFunctions.js` has
always done for segment context and footer tabs. Core contains no registry, no manifest and no
customisation vocabulary; every call site reads as ordinary code.

| File / group | Change |
|------|--------|
| `globalFunctions.js` | Unchanged, and the model for the rest — its members were already replaceable properties of a plain object |
| `components/segments/segmentNotes.js` (+test) | New. The four note members that used to be prototype-patched methods — the notes list, a single note, note content, metadata notes — as plain functions of an explicit context object, so an implementation never needs `this`. Its internal calls go through the object too, so replacing a member replaces it for the members that call it |
| `components/segments/matchInfo.js` | New. The extra information shown beside a match; core renders none |
| `components/header/headerInterface.js` (+test) | New. Extra header links, replacing the class whose prototype was patched |
| `pages/CatToolInterface.js` (+test) | Was a class extending a synthetic-`this` base; now a plain object with the same single member |
| `hooks/useProjectTemplates.js` | Same, for the character counter preset applied to new templates |
| `components/segments/SegmentFooterTabMessages.js` (+test) | Convert to a function component; the notes tab renders `segmentNotes.getNotes(...)` instead of exposing four prototype-patchable methods |
| `components/segments/SegmentFooterTabMatches.js` (+test) | Convert to a function component; port `shouldComponentUpdate` to a `memo` comparator; extra match information comes from `matchInfo.getMatchInfoMetadata(...)`; two dead plugin hooks removed — see Notes |
| `components/segments/SegmentFooter.js` | Replace an inline synthetic-`this` object (a reimplementation of the old extend interface) with a direct call to the notes object |
| `components/header/Header.js` | Call the header object's member instead of the patched class |
| `utils/ComponentExtendInterface.js` (+test) | Deleted. Its `get props() { return this }` hack existed only so a plugin object could impersonate a React `this`; with plain functions of an explicit context there is nothing to impersonate |

Everything else a plugin touches — `utils/segmentUtils.js`, `utils/commonUtils.js`,
`actions/SegmentActions.js`, `api/getContributions/`, the glossary components — is **byte-identical
to before this branch**. Those points were already module or object properties and never needed
changing; only the four React-prototype ones did.

### Tooling and housekeeping

| File | Change |
|------|--------|
| `.eslintrc.js` | Guard **every** class component via anchored `no-restricted-syntax`, now at `error` since none remain. Add an override for plugin front-end sources (previously a fatal parse error under `yarn lint`) that bans assigning to a prototype, since there is no longer one to patch. Add an optional rule keeping deployment names out of core |
| `.eslint-private-names.example.json`, `.gitignore` | Documented example for the name list above; the real list is supplied locally and by CI, so no names are committed |
| `jest.config.js` | Drop the coverage threshold for a directory this branch no longer creates — jest fails a coverage run when a threshold path matches no files |
| `public/CLAUDE.md` | Record that no class components remain and that lint enforces it; document the helper-object pattern, including the capture-then-replace rule that lets two deployments wrap the same member |
| `package.json`, `yarn.lock` | Pin `@babel/core`/`@babel/eslint-parser` back to `^7.x`; stop ESLint from loading the jest-only babel config |
| 16 migrated components and tests | Restore Prettier formatting (formatting only) |
| `MateCat-docs`, `plugins/*` | Bump submodule pointers |

## Testing

Frontend `yarn test` passes in full: **427 suites / 4444 tests / 2 snapshots** green, and
`jest --coverage` exits clean with every configured per-directory threshold met.

Every converted file was checked against ESLint/Prettier baselines (no new errors introduced) and
went through at least one independent break/restore mutation test to confirm the existing suite
actually protects the converted behavior, not just that it passes. For the Wave 4d files, the
converted source was additionally diffed against the extracted original line-by-line and reviewed
by an independent reviewer before commit.

The new seam modules are fully covered: `segmentNotes.js`, `matchInfo.js`, `headerInterface.js` and
`CatToolInterface.js` all measure 100%. The two newly converted components measure 100%
(`SegmentFooterTabMessages`) and 98.3% statements / 99.2% lines (`SegmentFooterTabMatches`).

**Manual QA still outstanding — please read.** This work was done in a non-interactive environment
with no running Matecat instance, so no live-browser pass was performed on either the Wave 3
review-extended panel cluster or, more importantly, the **segment editor**
(`Editarea`, `SegmentSource`, `SegmentTarget`, `SegmentPlaceholderLite`, and now the notes and
matches footer tabs). Automated coverage for those files is exhaustive, but the segment editor is
the highest-traffic surface in the product and should get a human pass — open a job, type in the
editor, insert/split tags, use composition (IME) input, triple-click select-all, and open the
glossary/QA decorators — before this merges.

Because this changes both core and the deployment-specific frontend plugins, **QA should run
against at least one deployment that customises the segment editor**, checking the notes tab, the
extra match information, the header links and the character counter.

## AI Disclosure

Claude Code (Sonnet 5, Opus 5)

## Notes

**Base branch**: this PR targets `coverage-improvement-plan` rather than `develop`, because that is
this branch's actual immediate ancestor, and all commits on it belong to this migration; `develop`
sits hundreds of commits behind that branch via other still-unmerged, long-lived integration work,
so diffing against `develop` would pull in unrelated history. Merge order should follow
`coverage-improvement-plan` landing first, or this branch should be rebased once it does.

**Plugin-side changes are pushed and pinned.** Each deployment plugin has its own branch and its own
PR, and the submodule pointers here are advanced to those commits, so this branch is checkout-clean
and CI exercises the matching pair. Core and plugins must land together: a plugin that still patches
a prototype stops working silently, because there is no prototype to patch.

**This branch's history contains a design that was built and then removed.** An earlier pass put
these points behind a named extension-point registry with slots and capability flags. It was
rejected as too invasive for open-source core — deployment customisations are expected to disappear
eventually, so core should not grow vocabulary for them — and the later commits take it back out. It
also had a real defect: because a wrapper asked for *core's* default rather than the currently
installed implementation, two deployments wrapping the same point silently discarded one of the two.
Capture-then-replace composes instead. Reviewing the branch diff rather than commit-by-commit avoids
the detour; say so if the intermediate commits should be squashed away.

**Inherited red CI (one line of PHP), deliberately left alone.**
`SegmentTranslationIssueControllerTest` asserts `'Not Authorized, …'` while
`ProjectAccessValidator` throws `'Not authorized, …'`, so `ci-cd (Freddy)` fails on it. The
mismatch is present unchanged on `coverage-improvement-plan` and `expectExceptionMessage` is
case-sensitive, so that assertion cannot pass — it is inherited, not caused here.

This PR is frontend-only by design: **`lib/` and `tests/` have a zero-byte diff** against the base.
A one-line fix to the assertion was briefly on this branch and has been rebased out to keep that
boundary clean, which means the failure is still visible here. It belongs with the sentence-case copy
work, not with a class-to-hooks migration. The correct side is the source string —
CLAUDE.md's sentence-case rule applies, and the test should follow the code.

**Incidental fixes bundled into the refactor** (each was a previously-untested bug naturally fixed by
the hooks conversion, not a separate scope-creep change): a `mousedown` listener leak in
`LanguageSelectorSearch`; an XSS-sanitization gap plus a link-rewrite crash in `JobMetadataModal`;
two `addListener`/`removeListener` reference-mismatch leaks in the review-extended panel cluster
(`ReviewExtendedPanel`, `ReviewExtendedIssuesContainer`). All four are covered by new regression
tests added alongside the fix.

**`TagEntity` and the two update bugs it exposed.** 36 of this component's 39 existing tests called
`new TagEntity(...)` and spied on instance methods, so the "safety net" for the conversion would
have disappeared at exactly the moment it was needed. The suite was therefore rewritten as
behaviour tests against the *unchanged* class first, then mutation-tested (12 seeded defects, 12
killed, including hooks-specific ones like non-merging state updates and stale-closure reads in a
store handler), and only then was the component converted. Because the tests no longer touch
instance internals, the conversion needed no `instanceRef` bridge and is plain hooks.

Two real bugs surfaced, each fixed in its own revertible commit with a test pair pinning both
directions: the update guard compared `prevProps.entitykey` (lowercase `k`) against
`props.entityKey`, so it was always true and every tag in the segment re-measured its overflow on
every update; and the mount measurement returned `null` rather than `false` when there was nothing
to measure, so every tag committed a second, pointless render on mount. The latter is the cascade
that produced "maximum update depth exceeded" during Wave 4d.

**Behaviour changes worth calling out**, all deliberate:

- `SegmentFooter`'s "does this segment have notes" check now asks the real note implementations,
  including any replacement a deployment has installed. It previously asked a synthetic stand-in
  that supplied its own cheaper versions of the same pieces, so it could disagree with what the tab
  actually rendered.
- `SegmentFooterTabMessages` intentionally has **no** `memo` comparator. The
  `shouldComponentUpdate` it replaces tested `this.props.note` — singular, never a prop — so its
  first clause was always true and the component always re-rendered. A working comparator would be
  a new optimisation, not a port of the old one.
- Two dead plugin hooks were deleted from the matches tab: a regexp field nothing ever assigned,
  and a match-processing callback that was an identity function behind an always-true guard.
- A notes entry whose payload is an object still renders nothing. The old loop returned its markup
  into the void, so this has never displayed; it is preserved as-is, with a comment, rather than
  silently starting to render new content inside a refactor.
- The size-restriction entry is filtered out of a segment's metadata before the notes object sees
  it, since the editor shows that on its own and it was never a note. No visible change.
- Links rendered inside a note carry `rel="noreferrer"`.

**Not addressed here.** `components/header/Header.js` dereferences `filterProjectsRef.current
.currentUser` during render while the guard beside it doesn't check that the filter is shown; this
looks like a live crash but is unconfirmed and unrelated to the migration, so it is left alone.
`SegmentFooter.js` also carries a `setTabLabel` callback that has been dead since it arrived in
#4368 — flagged, not removed, since it isn't this branch's code.

---
  - https://app.asana.com/0/0/1217582259934675


riccio82:
Superseded by the split. The work landed on develop as fourteen PRs:

#4806, #4813, #4814, #4815, #4816, #4817, #4818, #4822, #4824, #4835, #4840, #4842, #4844, #4847

plus the four plugin PRs — uber #74, airbnb #124, translated #68, aligner #329 — on their own
masters, with develop's submodule pointers moved to match.

develop now has zero React class components under `public/js`, and #4847 replaced the
per-directory eslint allowlist with a tree-wide ban, so the class form is a lint error anywhere
under `**/js/**` rather than only in the directories the ratchet had reached.

Closing rather than merging: this branch is stale by 236 files and holds nothing unlanded. Of the
three files unique to it, `public/js/components/common/TextField.js` and
`public/js/components/segments/SegmentPlaceholderLite.js` (with their tests) are imported by
nothing, and `segmentNotes.test.js` targets a version of `segmentNotes.js` that differs from the
merged seam by 104/-85 lines, so it does not apply as written.

PR #4784, opened 2026-08-25
✨ feat(job-metadata): persist MT settings per job with a project-metadata fallback

mauretto78:
## Summary

The MT tuning settings — DeepL formality, Lara style and glossaries, the MMT and Intento options,
the MT application threshold — were written to `project_metadata` at creation time only, so they
could never be changed afterwards. They are now stored per job, which also lets two target
languages of the same project hold different values. Prerequisite for per-job Lara instructions.

Production cannot be migrated (`project_metadata` holds billions of rows), so every read goes
through a fallback chain: **job metadata → project metadata → null**. The project scope is therefore
permanent, not a deprecation window.

The write path is **copy-on-write**: creation keeps writing these keys to `project_metadata` as the
project's base value, and a `job_metadata` row appears only where the owner actually overrides one.
That is what makes the change revertible — revert it and every project still resolves to a correct
value rather than to the engine default. It also means both scopes can hold the same key, and the
job one wins whenever it is present.

Backend only, no JS touched. Frontend integration notes follow as a comment on this PR.

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Changes

| File | Change |
|------|--------|
| `lib/Model/Jobs/JobSettingsResolver.php` | **new** — resolves a setting job → project → null; `resolveMany()` costs two lookups for a whole set instead of two per key |
| `lib/Model/Jobs/JobsMetadataMarshaller.php` | the 11 new keys and their casts, mirroring `ProjectsMetadataMarshaller` so both scopes return identical types; `mtSettings()` and `propagatedOnSplit()` single-source the key lists |
| `lib/Model/ProjectCreation/JobCreationService.php` | unchanged from `develop` — creation writes no MT row per job at all under copy-on-write |
| `lib/Model/ProjectCreation/ProjectMetadataService.php` | keep writing them at project level as the creation-time base value, with the scalar guard the per-job write used to own — `!empty()` would drop `mmt_ignore_glossary_case = false` and a threshold of `0`, which are answers rather than absences |
| `lib/Utils/Engines/{DeepL,MMT,Lara,Intento}.php` | read through the resolver; it also absorbs the `pid` vs `id_project` split that used to null half the settings depending on the caller |
| `lib/Utils/AsyncTasks/Workers/Analysis/TMAnalysisWorker.php` | carry `job_password` into the engine config so the engines can address a job row |
| `.../Analysis/TMAnalysis/Service/EngineService.php` | `lara_style` via the resolver; add the missing cache TTL to the adjacent `enable_mt_analysis` read, which ran uncached once per analysed segment |
| `lib/Utils/AsyncTasks/Workers/Analysis/FastAnalysis.php` | resolve the threshold per job into the existing per-`(id_job, password)` cache, with the project row as the second leg and the value `main()` materialised as the last resort. That read runs in the publish loop, outside any transaction, so ProxySQL can answer it from a lagging replica; with the project leg switched off such a miss fell through to a hard-coded `85` |
| `lib/Controller/API/App/{GetContributionController,SetTranslationController}.php` | resolve `lara_style` and the threshold |
| `lib/Controller/API/V3/MetaDataController.php` | report the engine keys under `mt_extra` on the job scope too, via one shared engine-key helper |
| `inc/validation/schema/job_metadata.json` | accept the new keys, with real value schemas (enums, 0–100, typed glossary arrays); require `key` and `value` on **all 16** branches, including the 5 that predate this PR — without it `[{"key":"deepl_formality"}]` stored the four-character string `"null"` past the enum, and `[{}]` was a 500 rather than a 400. Bound the glossary arrays: `[]` was writable and the Lara SDK forwards it verbatim as `"glossaries": []` |
| `lib/Model/Jobs/{JobDao,MetadataDao,JobCredentialCacheInvalidator}.php` | move `job_metadata` onto the new password when a job password is rotated — `JobDao` never did, so the rows were stranded under a credential nothing resolves any more. Evict the **per-key** reads too, not only the bulk one: `MetadataDao::get()` binds the key as well, so those entries hash elsewhere and answered the retired password for the full 86400s TTL. The metadata DAO is now a required constructor argument, since nullable-plus-`?->` made omitting it a silent no-op |
| `lib/Model/Jobs/JobsMetadataMarshaller.php` | one `DEFAULT_MT_QUALITY_VALUE` constant replacing six hard-coded literals that disagreed — 85 in the analysis, **86** in the editor, for the same unset key; both are consumed as `100 - $value`, so the penalty was 15 in one and 14 in the other. Add the four keys creation writes but a split silently dropped: `dialect_strict`, `mandatory_issues`, `public_tm_penalty`, `tm_prioritization` |
| `lib/Model/JobSplitMerge/JobSplitMergeService.php` | propagate/delete from `propagatedOnSplit()` instead of a hardcoded 3-key list, so a new key cannot silently vanish from the chunks a split creates; pass the metadata DAO to the credential sweep, which this third call site omitted |
| `tests/**` | the resolver's fallback contract, the copy-on-write creation path on both scopes, the split/merge propagation, the endpoint shape, and every accepted/rejected payload. Two real-SQL regression tests for the per-key cache eviction, both verified to fail before the fix |

## Testing

PHPStan is clean at level 8 with no baseline.

Rebased onto `develop` (580 commits). Seven conflicts, all in files this PR rewrites; the
substantive one is `TranslatorsModel::changePassword()`, which `develop` moved into
`$database->transaction()` with the sweep registered via `onCommit(critical: true)`. Two callers also
had to move onto `develop`'s single `MetadataDao::destroyCache(MetadataStruct)` door, since the
per-address eviction methods became private.

PHPUnit locally: the remaining failures are environmental and pre-existing on `develop` in my
container — 4 assert that ActiveMQ is *unreachable* while `matecat-amq` is up, 2 depend on a locally
stale built `lib/View/index.html` (CI rebuilds it), and 4 `UserDao`/`SignupModel` confirmation-token
cases fail because my local test database still has `users.confirmation_token` as `varchar(50)` where
the checked-in schema declares `varchar(255)`, so `develop`'s scoped token is truncated on insert.
None of the three groups touches this PR's code.

Not yet done: the Bruno `dev` end-to-end pass (create a project with the settings → assert the rows
land in `project_metadata` → change one over the API → assert a `job_metadata` row appears and
`getContribution` reflects it → `DELETE` the key and confirm it falls back to the project value).

## AI Disclosure

Claude Code (claude-opus-5)

## Notes

**Release order.** Copy-on-write relaxes this a lot compared with the first version of the PR: the
`project` scope still carries the value for every project, old and new, so today's UI keeps showing
the MT settings panel correctly on a freshly created project. What still needs the frontend is the
owner's *edits* — those are written to `job` only, so until the UI reads `job` first and falls back to
`project`, a change made through the settings panel will not be reflected back. The frontend change is
harmless against today's backend (`job.mt_extra` is simply an empty object), so it can land first,
together, or shortly after. No frontend PR is open yet.

Two keys deliberately stay project-scoped: `enable_mt_analysis`, because the analysis the customer
was quoted for was priced on it, and `mmt_activate_context_analyzer`, which only
`MMT::syncMemories()` reads, once, at creation, from a project row — changing it later could not
have any effect.

Fixed while here, all pre-existing: the four keys that did not survive a job split
(`dialect_strict`, `mandatory_issues`, `tm_prioritization`, `public_tm_penalty`) are now in
`propagatedOnSplit()`, and a test asserts the list covers *every* declared key so the next one added
cannot repeat the bug. The MT-threshold default is unified on 85 — **note this is a user-visible
shift**: the editor previously defaulted to 86, so a project with no stored threshold now applies an
MT penalty of 15 rather than 14, matching what the analysis already priced.

Still open, deliberately not fixed here: on merge only the first chunk's `job_metadata` rows survive,
so an MT edit made on a later chunk only is lost. Under copy-on-write that now degrades to the
project's creation-time value rather than to the engine default, which is why it is not urgent.

There is no way to express "use no glossary on this job while the project has one". Under
copy-on-write `DELETE` means "inherit the creation-time value", and the alternative — an empty-string
sentinel — is not uniform across the eleven keys (an empty row decodes to `0` for the threshold and
`false` for `mmt_ignore_glossary_case`, neither of which means "none"). Worth a separate ticket if the
product wants it.

Frontend integration notes for @riccio82 and @piedicianni follow as a comment below.

---
  - https://app.asana.com/0/0/1209716725975762


mauretto78:
## Frontend integration notes — MT settings moved from the project to the job

@riccio82 @piedicianni — this PR changes **where** the MT settings live, not which endpoints you
call. `getJobMetadata` and `updateJobMetadata` stay exactly as they are; what changes is which
scope of the response holds the value.

### The rule

`GET /api/v3/jobs/{id_job}/{password}/metadata` still returns `{project, job, files}`. For the 11
MT settings below:

> **read `job` first, fall back to `project`.**

> **Updated after review** — the backend now writes **both** scopes. The rule above is unchanged;
> what changed is that `project` is always populated, which makes this a much softer landing for you.

- **`project` always has them**, for every project, old and new. Those rows can never be migrated —
  `project_metadata` holds billions of rows in production — so the project scope is a permanent
  fallback, not a deprecation window. Don't plan to remove it.
- **`job` has them only where the owner overrode one**, per target language. That is the only scope
  later edits are written to, so `job` is the answer whenever the key is present there.
- **Both scopes can hold the same key**, and `job` wins. This is the one change from my original
  notes, which said "never both" — a plain `project`-then-`job` spread is still exactly right, so no
  code changes from what is described below.

### Where each key lives, and its type

Engine configuration parameters are nested under `mt_extra` **on both scopes**. The application
threshold is not an engine parameter, so it sits flat on the scope — the same place it does today.

| key | position | type as returned |
|---|---|---|
| `deepl_formality` | `mt_extra` | `string` |
| `deepl_id_glossary` | `mt_extra` | `string` |
| `deepl_engine_type` | `mt_extra` | `string` |
| `lara_style` | `mt_extra` | `string` |
| `lara_style_guideline_id` | `mt_extra` | `string` |
| `lara_glossaries` | `mt_extra` | `array<string>` (already decoded) |
| `mmt_glossaries` | `mt_extra` | ⚠️ `string` — a JSON string like `"[101]"`, **not** an array |
| `mmt_ignore_glossary_case` | `mt_extra` | `boolean` |
| `intento_provider` | `mt_extra` | `string` |
| `intento_routing` | `mt_extra` | `string` |
| `mt_quality_value_in_editor` | flat on the scope | `number` (int) |

**Two engine parameters deliberately stay project-only — don't look for them on `job`:**

- `enable_mt_analysis` — the analysis the customer was quoted for was priced on it.
- `mmt_activate_context_analyzer` — only `MMT::syncMemories()` reads it, once, at project creation,
  from a project row. Changing it afterwards could not have any effect, so it is not made editable.

`subfiltering_handlers` is unrelated to this change: it already lives on `job` (and is mirrored on
`project` for the analysis). Keep reading it from `job` as you do now.

### Two guarantees that make a plain merge correct

1. **`mt_extra` is always present on both scopes**, as an empty object when there are no rows. So
   `project.mt_extra.lara_style` yields `undefined` rather than throwing — the existing
   non-optional accesses (e.g. `LaraStyles.js:17`) won't start crashing on new projects. And since
   `project` is now always populated, they keep returning a real value on a new project too.
2. **Both scopes un-marshall these 11 keys to identical types**, and the backend **omits absent keys
   entirely** instead of emitting `null`. That is what makes a spread merge safe: a later spread
   can never overwrite a present project value with an absent job one.

### Reading — the main integration point

`public/js/pages/CatTool.js` (~L625-655), where `jobMetadata` becomes the settings-panel template:

```js
// before
mtQualityValueInEditor: jobMetadata.project.mt_quality_value_in_editor,
mt: {
  ...prevTemplate.mt,
  extra: {
    ...(jobMetadata.project.mt_extra ?? {}),
  },
},

// after
mtQualityValueInEditor:
  jobMetadata.job.mt_quality_value_in_editor ??
  jobMetadata.project.mt_quality_value_in_editor,
mt: {
  ...prevTemplate.mt,
  extra: {
    ...(jobMetadata.project.mt_extra ?? {}),
    ...(jobMetadata.job.mt_extra ?? {}),   // job last → job wins, key by key
  },
},
```

### The other read sites need the same fallback

These read `project.mt_extra` straight off the payload rather than through the template, so the
merge above does **not** cover them:

| file | line |
|---|---|
| `components/segments/ToolbarFeatures/Ai/LaraStyles.js` | 17 |
| `components/segments/utils/translationMatches.js` | 313, 340 |
| `components/segments/SegmentFooterTabAiFeedback.js` | 55 |
| `components/segments/SegmentFooterTabAiAlternatives.js` | 217 |
| `components/segments/SegmentFooterTabLaraStyles.js` | 64 |
| `components/settingsPanel/Contents/MachineTranslationTab/MTGlossary/MTGlossary.js` | 230 |
| `components/settingsPanel/Contents/MachineTranslationTab/LaraGlossary/LaraGlossary.js` | 62-63 |
| `components/settingsPanel/Contents/MachineTranslationTab/DeepLGlossary/DeepLGlossary.js` | 212 |

A small shared helper is probably worth it, e.g.
`const mtSetting = (jobMetadata, key) => jobMetadata?.job?.mt_extra?.[key] ?? jobMetadata?.project?.mt_extra?.[key]`.

### Writing — nothing changes

Read is v3, write is the `/api/app/` alias. That asymmetry already exists and is unchanged:

| | endpoint | existing helper |
|---|---|---|
| read | `GET /api/v3/jobs/{id_job}/{password}/metadata` | `api/getJobMetadata/getJobMetadata.js` |
| write | `POST /api/app/jobs/{id_job}/{password}/metadata` | `api/updateJobMetadata/updateJobMetadata.js` |
| clear | `DELETE /api/app/jobs/{id_job}/{password}/metadata/{key}` | — |

`updateJobMetadata.js` already posts the right shape (`[{key, value}, …]`) to the right place — the
new keys just need adding to its `paramsData` map. Authorization is unchanged: login + chunk
password + **team access**, so only a member of the team owning the project can write, exactly as
for today's job settings.

### Accepted values

Validated by JSON Schema (`inc/validation/schema/job_metadata.json`). A mismatch is **rejected with
a schema error**, not silently ignored — so please send the right types:

| key | accepted |
|---|---|
| `mt_quality_value_in_editor` | integer `0`-`100` — an **integer**, not a string |
| `deepl_formality` | `default` \| `prefer_less` \| `prefer_more` |
| `deepl_engine_type` | `prefer_quality_optimized` \| `latency_optimized` |
| `lara_style` | `faithful` \| `fluid` \| `creative` |
| `lara_glossaries` | array of non-empty strings |
| `mmt_glossaries` | array of integers |
| `mmt_ignore_glossary_case` | boolean |
| `deepl_id_glossary`, `lara_style_guideline_id`, `intento_provider`, `intento_routing` | non-empty string, ≤255 chars |

### Three gotchas

1. **`mmt_glossaries`: you POST an array, you GET a JSON string.** The controller JSON-encodes
   arrays on write, and this key is read back as the raw string the engine parses itself. So
   `POST {"key":"mmt_glossaries","value":[101]}` → `GET … "mmt_glossaries": "[101]"`.
   `lara_glossaries` is symmetric (array in, array out). This mirrors the existing project-scope
   behaviour, so it is not new — but heads up that `MTGlossary.js:230` calls
   `project.mt_extra.mmt_glossaries?.some(…)`, i.e. an array method on a string. Whatever makes
   that not blow up today, it is worth parsing explicitly while you are in there.
2. **There is no "set to empty".** Empty strings are rejected on purpose: an empty job row is an
   *answer*, not an absence, and would shadow the project fallback with a value the engine cannot
   use. Empty **arrays** are rejected now too (`lara_glossaries: []`, `mmt_glossaries: []`), for the
   same reason — the Lara SDK forwards `[]` verbatim rather than omitting the parameter. To clear a
   setting, call the `DELETE` endpoint; the job then falls back to the project value. There is
   currently **no way to say "use no glossary on this job while the project has one"** — flag it if
   the UI needs it and we'll add it as its own change.

   Also note both `key` and `value` are now mandatory on every item: `[{"key": "deepl_formality"}]`
   with no value is rejected with a schema error rather than quietly storing a bogus value.
3. **Settings are per chunk**, keyed by `(id_job, password)`. Two target languages of the same
   project can now hold different values, and a job split gives each chunk its own copy. Don't
   cache one chunk's MT settings and reuse them for another.

### Release order — relaxed

This is the part the review changed most, in your favour. **The hard ordering constraint is gone.**
Because `project` is still written at creation, today's UI keeps showing the MT settings panel
correctly on a newly created project — it no longer comes up empty.

What still needs the frontend change is the owner's **edits**: those are written to `job` only, so
until the UI reads `job` first and falls back to `project`, a change made in the settings panel
won't be reflected back on reload. So the frontend change can land first, together, or after — after
is now merely a feature gap rather than a visible regression.

Ostico (changes requested):
## Review: persist MT settings per job with a project-metadata fallback

The read layer in this PR is the right design and I would keep almost all of it. `JobSettingsResolver`,
the write schema, the ten rewired read sites, the password-rotation fix and the split/merge
propagation are all solid and all of them survive the change requested below unchanged.

The request is narrower: **keep writing the MT keys to `project_metadata` at creation, and write a
job row only when a setting is actually overridden.** That removes two pieces of the creation path
and, more importantly, makes this change revertible. Details and the exact keep/remove list follow.

Two notes before the detail. Every line number below is against the current branch head and will
drift once you rebase. And I merged `origin/develop` into a scratch branch to check whether the
rebase resolves any of this on its own: **it resolves 1.1 and nothing else** — 1.2, 1.3 and 1.4 all
survive the merge unchanged. Evidence is in each item.

---

### 1. Blockers (independent of the design question)

**1.1 — The branch is unmergeable and its CI evidence has expired.**
544 commits behind `develop`, 1 ahead, `mergeable: CONFLICTING`. The conflicts are in
`GetContributionController.php`, `MetaDataController.php`, `TranslatorsModel.php`,
`TMAnalysisWorker.php` and `tests/unit/Core/Workers/Analysis/FastAnalysisTest.php` — the same files
this PR rewrites. The suite on the branch reports 91 errors and 25 failures, dominated by
`MethodCannotBeConfiguredException: Trying to configure method "begin"` at `JobDaoTest.php:41`,
which is fallout from the transaction API landing on `develop` after this branch forked. Nothing
here can be judged until it is rebased.

I ran the merge to see what it costs. Five conflicts, all in files this PR rewrites, and
`TranslatorsModel.php` is the one to watch — it is one of only two call sites that correctly passes
the metadata DAO, so its resolution decides whether 1.3 gets worse. The good news is that on the
merged tree PHPStan reports `[OK] No errors` across `JobCredentialCacheInvalidator`, `MetadataDao`,
`JobSettingsResolver`, `JobSplitMergeService`, `JobCreationService` and `ProjectMetadataService`:
none of develop's newer transaction rules flag anything this PR added.

**1.2 — Password rotation does not evict the per-key metadata cache.**

To be fair about provenance: this is not a regression you introduced. On `develop`,
`JobCredentialCacheInvalidator` takes three arguments — `JobDao`, `ChunkReviewDao`, `ProjectDao` —
and evicts no job metadata at all, even though `develop` already writes seven metadata keys per job.
The gap predates this PR. What this PR does is add the eviction mechanism while also introducing the
per-key read path that the new eviction does not cover, so the fix belongs here by virtue of being
where the mechanism now lives.

The mechanism: `sweepAfterJobPasswordRotation()` calls `MetadataDao::destroyCacheByJobAndPassword()`
(`:81`), built from `_query_metadata_by_job_password` bound with **two** values,
`(id_job, password)`. But every read through `JobSettingsResolver::resolve()` goes via
`MetadataDao::get()` (`:103`), which caches under `_query_metadata_by_job_password_key` bound with
**three**, `(id_job, password, key)`. Different statement, different bind set, so the hashed cache
keys can never collide. The per-key entries survive the rotation for the full TTL of 86400 seconds,
keyed on the retired password.

`destroyCacheByJobAndPasswordAndKey()` already exists at `:118`; it is simply never called from the
sweep. Please call it per key there, or add a bulk equivalent.

I confirmed this against the merged tree, not just the branch: both methods and both query constants
are still distinct after merging `develop`, so the rebase does not close it.

Consequence: any flow that returns a job to a previously used password reads its pre-rotation
settings.

**1.3 — `JobSplitMergeService::sweepCredentialCaches()` does not pass the `JobsMetadataDao`.**
Lines 176-182 on the branch as it stands; lines 170-174 after merging `develop`, where the call site
still constructs the invalidator with three arguments. `ChangePasswordController:153` and
`TranslatorsModel:339` both pass it; this third call site does not, so split and merge — which
rotate passwords — skip metadata eviction entirely. Given 1.2 the parameter should probably become
non-nullable so this cannot be omitted again.

**1.4 — The write schema never requires a value.**
All 11 new branches in `inc/validation/schema/job_metadata.json` set `additionalProperties: false`
and constrain `key` and `value`, but none declares `required`. So `[{}]` validates, and so does
`[{"key": "deepl_formality"}]` with no value at all. `JobMetadataController:108` then coerces the
missing value to the four-character string `null`, which `unMarshall()` hands to DeepL as the
`formality` parameter — bypassing the very enum this PR added. Please add
`required: ["key", "value"]` to each branch. Still zero occurrences of `required` in the file after
merging `develop`.

---

### 2. Requested change: write to the project at creation, to the job on override

#### 2.1 The stated reason for the current approach does not hold

`JobCreationService:239-242` and `ProjectMetadataService:88-96` both justify the eager move with the
same claim — that a second copy would go stale as soon as a job edits the value, and that reads
would have no way to tell which copy is current.

Reads have a very explicit way to tell. `JobSettingsResolver::resolve():58-71` returns the job value
whenever a row exists and only then consults the project; `resolveMany():95-106` seeds the map from
the project and overlays the job rows on top. Precedence is defined in code, in one place, and it is
already exactly copy-on-write semantics at the read layer.

The staleness hazard also has no code path today. I checked every write to `project_metadata` in
`lib/` and `plugins/`: there are exactly three — `ProjectMetadataService:153` and
`ProjectManager:285`, both at creation, and `ContextUrlController:168`, which only writes
`context_url`. Nothing mutates a project-scoped MT setting after creation, so a project copy that is
written once and never updated cannot go stale. It is a frozen creation-time snapshot, which is
precisely what a copy-on-write base value should be.

Finally, the same file already does the thing the comment forbids: `ProjectMetadataService:116-124`
deliberately writes `subfiltering_handlers` to *both* project and job metadata, with a comment
explaining that the project copy exists so analysis need not reach the jobs DAO. "One scope must own
the value" is not a principle this codebase holds.

#### 2.2 The decisive argument: this change is currently not revertible

`ProjectMetadataService:98-114` skips the MT keys and then unsets the threshold, so for every
project created while this is live the values exist **only** in `job_metadata`.

Revert the PR — for any reason, at any point — and every one of those projects silently loses DeepL
formality, Lara style, the glossaries and the MT threshold, falling back to engine defaults. There
is no migration in the 34 changed files, in either direction. Recovery would mean backfilling
`job_metadata` into `project_metadata` over the affected id range, and for a multi-target project
you would have to pick which job's value to promote — an arbitrary choice made under incident
pressure.

Under copy-on-write this problem does not exist. `project_metadata` still holds the creation-time
value for every project; revert the code and reads simply return to the project scope. The only loss
is the overrides made during the window, and every job lands on a correct value rather than an
engine default.

#### 2.3 Three states are already representable, using a mechanism this PR relies on

The obvious objection to copy-on-write is the reset case: the project says
`deepl_id_glossary = 'abc'`, the user wants *this job* to use no glossary, and deleting the job row
means "inherit `'abc'`" rather than "none".

That case is already covered. `job_metadata.value` is `text NOT NULL`, and `''` is a perfectly valid
non-null value, so storage distinguishes three states:

| state | representation | meaning |
| --- | --- | --- |
| no row | — | inherit from project |
| row with `''` | `value = ''` | explicitly none |
| row with content | `value = 'abc'` | override |

And the read path already honours it. `unMarshall()` returns `(string)$struct->value`, so `''`
decodes as `''` rather than null; `resolve():60` short-circuits on `$struct !== null` regardless of
what the row contains; and `resolveMany():105` assigns `$resolved[$struct->key] = $struct->value;`
with no emptiness test, so an empty job value overwrites a non-empty project value.

This is the same mechanism the PR already depends on — `JobCreationService:264-266` says an empty
row "would shadow the project-metadata fallback", which is exactly the override semantics
copy-on-write needs.

Two consequential notes:

- The write schema currently blocks a user from *setting* that state: `deepl_id_glossary`,
  `lara_style_guideline_id`, `intento_provider` and `intento_routing` are `"type": "string",
  "minLength": 1`. If explicit-none should be settable through the API, relax `minLength` on those
  four. This does not affect internal writes — the schema validates the endpoint, not the column.
- `lara_glossaries` and `mmt_glossaries` are arrays with no `minItems`, so `[]` already expresses
  "no glossary" for them today.

#### 2.4 Storage, stated honestly

This is the weakest argument for the change and I do not want to oversell it. Only the selected
engine's keys are ever written, so the 11 keys are far sparser than they look. On the development
database the moved keys occupy 971 rows across 1146 projects — 0.85 rows per project, not 11 — and
at 1.079 jobs per project the eager write turns that into roughly 1048 rows: about 77 extra rows out
of ~10,375, well under 1%. Those proportions want confirming against production volumes before
anyone leans on them.

It only becomes material at the tail: a 50-target project later split four ways is 200 chunks, so
roughly 800 rows where copy-on-write writes 4. Worth noting, not worth deciding on.

#### 2.5 The argument against this request, for completeness

Copy-on-write makes the two-scope fallback permanently load-bearing, and it changes how a mistake
fails. A future read site that forgets `JobSettingsResolver` and queries `project_metadata`
directly returns nothing under the eager write — it breaks visibly, in the first test. Under
copy-on-write it returns the project value, correct for every job *except* the overridden ones:
silently wrong MT output for exactly the users who bothered to customise.

That is a real cost and I want it on the record. It is a code-review discipline problem rather than
a correctness barrier, and I do not think it outweighs an unrevertible migration — but if you
disagree, this is the argument to make, not the staleness one.

Note also that the fallback is permanent either way: there is no backfill for existing projects, so
`resolve()` must keep consulting the project scope for the entire pre-PR population regardless of
which design ships.

---

### 3. What to keep, what to remove

#### Keep unchanged — all of this is required identically under copy-on-write

| Component | Note |
| --- | --- |
| `lib/Model/Jobs/JobSettingsResolver.php` | The whole class. `resolveMany()` is already written as a project-then-job merge. |
| `inc/validation/schema/job_metadata.json` | Keep, plus the `required` fix from 1.4. |
| `lib/Controller/API/App/JobMetadataController.php` | This is the override write path copy-on-write needs. No new endpoint required. |
| `lib/Model/Jobs/JobsMetadataMarshaller.php` | `mtSettings()`, `propagatedOnSplit()`, `unMarshall()`. |
| The ten rewired read sites | `DeepL`, `Lara`, `MMT`, `Intento`, `GetContributionController`, `SetTranslationController`, `MetaDataController`, `FastAnalysis`, `EngineService`, `TMAnalysisWorker`. |
| `MetadataDao::updateJobPassword()` + `JobDao:238,249` | This fixes a pre-existing bug: rotation already stranded rows for the seven keys `develop` writes eagerly. Needed regardless. |
| `JobCredentialCacheInvalidator` | Keep, with the 1.2 and 1.3 fixes. |
| Split/merge propagation | Keep verbatim. The existing `if (!empty($_data))` guard at `JobSplitMergeService:511` is already copy-on-write behaviour, and it is still required: chunks get fresh passwords, and `job_metadata` is keyed by `(id_job, password)`, so an override must be copied forward or it vanishes. |

#### Remove

| Location | What |
| --- | --- |
| `JobCreationService.php:229` | The `array_merge(..., self::buildMtSettingsMetadata($projectStructure))`. |
| `JobCreationService.php:234-276` | `buildMtSettingsMetadata()` and its docblock, in full. |
| `ProjectMetadataService.php:98` | `$jobScopedKeys = array_flip(JobsMetadataMarshaller::mtSettings());` |
| `ProjectMetadataService.php:101-103` | The `if (isset($jobScopedKeys[$extraKey])) { continue; }` guard — keep the loop, drop the guard. |
| `ProjectMetadataService.php:110-114` | The loop that unsets the job-scoped keys and the threshold. |

`saveJobsMetadata()` itself stays exactly as it is on `develop` — lines 192-228 and 231 predate this
PR and already write seven keys per job.

#### Change

1. **Port the scalar guard.** `ProjectMetadataService:105` uses `if (!empty($engineValue))`, which
   silently drops `0` and `false` — so `mmt_ignore_glossary_case = false` and a threshold of `0`
   never persist. `buildMtSettingsMetadata:268` got this right with
   `if (!is_scalar($value) || $value === '') { continue; }`. Please move that guard into the project
   write so the fix is not lost along with the method.
2. **Teach the engines about explicit-none.** The engines currently test `!== null`
   (`Intento.php:186` is `if ($customProvider !== null)`), so an empty-string row would be forwarded
   as an empty parameter rather than omitted. One added `&& $x !== ''` per site, across `DeepL.php`,
   `Lara.php`, `MMT.php` and `Intento.php`.
3. **Rewrite the two comments** at `JobCreationService:234-243` and `ProjectMetadataService:88-96`
   whichever way this lands. As written they assert a hazard that has no code path and are
   contradicted 20 lines below by the `subfiltering_handlers` dual-write.

---

### 4. Open question

Can `FastAnalysis`'s `$this->db()` resolve to a read replica?

`resolveJobAnalysisMetadata():978-997` passes `null` as the project id, disabling the fallback leg,
and relies instead on `$projectMtQualityValue`, which `main()` reads from the master inside a
transaction at lines 516-527 and 555. The final expression at line 996 is
`$jobMtQualityValue ?? $projectMtQualityValue ?? false`.

Under this PR the project row no longer exists for new projects, so line 555 falls to its hard-coded
`?? 85`. If that job-row read can hit a lagging replica seconds after creation, the user's chosen
threshold is silently replaced by 85 with no backstop. Under copy-on-write the master-read project
value remains the backstop. Worth confirming either way.

---

### Summary

Blocking: 1.1 rebase, 1.2 per-key cache eviction, 1.3 the missing DAO argument, 1.4 schema
`required`. The rebase resolves 1.1 alone — I verified 1.2, 1.3 and 1.4 all persist on the merged
tree, so please treat them as separate work rather than expecting the merge to absorb them.

Requested: move the MT keys back to a creation-time project write and write job rows only on
override — roughly 71 lines removed from the creation path, the revert path restored, and no change
at all to the resolver, the schema, the controller, the read sites or the split logic.

mauretto78:
## Re: persist MT settings per job with a project-metadata fallback

Thanks — this is a good review and the central request is right. **Accepted and done: the creation
path is copy-on-write, and all four blockers are fixed.** Pushed as `c06475c7`; the branch is
`MERGEABLE` again. Details below, plus five corrections and seven things the review missed.

### Section 1 — all four confirmed, all four fixed

**1.2** is the sharpest of them and I'd missed it entirely. Confirming the mechanism you describe,
with one addition: it isn't only the SQL string and the bind count that differ. `AbstractDao:378-381`
derives the Redis *hash-set* name from the calling method plus the bind values too, so a `get()` read
lands in `MetadataDao::get-<idJob>:<password>:<key>` while the eviction deletes from
`MetadataDao::getByJobIdAndPassword-<idJob>:<password>`. Two independent reasons the keys can't
collide.

The fix landed on top of your own work rather than beside it. `develop` had already funnelled
job-metadata eviction through one public door, `MetadataDao::destroyCache(MetadataStruct)`, which
names all three addresses a row answers — and made the per-address methods private, so the branch's
existing call was a fatal error the moment the merge landed. So there is now a
`destroyCacheForCredential($idJob, $password)` built on that: it iterates
`JobsMetadataMarshaller::cases()`, and the sweep calls it from inside the existing `credentials()`
loop, which already covers both the old and the new password. One line at the call site.

One deliberate departure from going through `destroyCache()` per key: the credential-bound address
doesn't depend on the key, so naming it eighteen times would triple the round trips for nothing. It
is named once, outside the loop, and only the two key-bound addresses are named per key. That is its
own commit (`c06475c7`) — a merge sweeps every chunk it folds in, so a fifty-chunk merge was issuing
a few thousand redundant deletes.

**1.3** confirmed, and worse than "skips metadata eviction": the merge path *does* rotate the rows
(`JobDao:946-948` → `changePassword()` → `MetadataDao::updateJobPassword()`), so that call site
strands *every* metadata cache entry, not just the per-key ones. Agreed on making the parameter
non-nullable — and the reason it has to be non-nullable rather than just documented is the `?->` at
`:61`: nullable-plus-null-safe-operator is what makes the omission silent instead of fatal.

**1.4** confirmed. One thing to add: the branches don't require `key` either. `[{}]` and
`[{"value": true}]` also validate, and then `JobMetadataController:107` passes `null` into
`MetadataDao::set(… string $key …)` — a **TypeError and a 500**, not a 400. So it's
`required: ["key", "value"]`, and on all **16** branches, not just the 11 new ones — the five
pre-existing ones have the same hole.

### Corrections to 1.1

Rebased — `mergeable: MERGEABLE`. But three things worth pinning down, because the rebase was not the rewrite the review implies:

1. **CI on the branch was already green** before I touched it — every check passing, including
   `ci-cd / Run tests` (6m09s) and Test-Guard, whose first run failed Layer 3 on `FastAnalysis` and
   whose later run, after the tests landed, was ✅ PASS. The "91 errors and 25 failures" figure isn't
   from this PR's CI; it looks like a local or merged-tree run. Worth flagging because it changes how
   the rebase gets scoped.
2. **`IDatabase::begin()/commit()/rollback()` are gone on `develop`, and no file this PR changes
   calls them.** I grepped `->begin()`, `->commit()`, `->rollback()`, `->onCommit(` and
   `->transaction(` across every PHP file in the diff: zero hits. So the transaction-API landing
   won't force any rewrite inside this PR's own code — it's confined to the conflicts.
3. **Seven conflicts, not five.** You missed `lib/Model/Jobs/JobCredentialCacheInvalidator.php` and
   `tests/unit/Core/Model/JobSplitMerge/SplitJobMergeTest.php`. The invalidator one matters, because
   `develop` renamed the `ChunkReviewDao` methods it calls.

And you were right that `TranslatorsModel` is the one to watch, though for a different reason than
the 4th argument: `develop` restructured `changePassword()` into `$this->database->transaction(…)`
with the sweep registered via `onCommit(fn() => …, critical: true)`. The DAO is threaded into that
scope rather than passed to a bare constructor call. The other notable resolution is
`GetContributionController`, where `develop` moved the `MateCatFilter` construction inside the
`!$concordance_search` block with an extra `$sourceContainsIcu` argument, so the branch's copy is
dropped rather than merged.

### Section 2 — accepted, and 2.1 is a fair hit

You're right that the staleness rationale doesn't hold, and right about why. I checked it the same
way you did and get the same three writes — `ProjectMetadataService:153`, `ProjectManager:285`,
`ContextUrlController:168` — plus zero in `plugins/`, zero callers of
`Projects\MetadataDao::delete()`, and one whole-project `DELETE` at `ProjectManagerModel:497` that is
project deletion, not a setting edit. Nothing mutates a project-scoped MT setting after creation. And
the `subfiltering_handlers` dual-write 20 lines below is the same file contradicting the comment,
as you say. Both comments are rewritten, and so is the `MetaDataController` docblock, which asserted
the same thing to API clients.

**2.2 is the deciding argument** and I'm not going to argue the other side of 2.5 — thanks for
putting it on the record anyway, it's the honest version of the trade. An unrevertible migration
isn't worth 46 lines of creation path (the actual figure once `buildMtSettingsMetadata()` and its
docblock came out).

### Pushing back on 2.3 — the empty-string mechanism

This is the one piece I did **not** implement. Two reasons.

**It isn't uniform across the 11 keys.** `unMarshall()` only routes 8 of them through the `(string)`
arm. The others:

| key | an empty row decodes to | meaning |
| --- | --- | --- |
| `mt_quality_value_in_editor` | `(int)''` → **`0`** | a real threshold of zero, not "none" |
| `mmt_ignore_glossary_case` | `(bool)''` → **`false`** | indistinguishable from the default |
| `lara_glossaries` | `json_decode('')` → **`null`** | reads as absent, but the row still shadows |

So the three-state table is only true for the string keys, and for the threshold it's actively wrong.

**And "explicitly none" is lost by this change, not gained by it.** Under the eager move, `DELETE` on
a new project's key leaves no row on either scope → engine default → "none". That's the state you
want, and it's already reachable through an endpoint that exists. Under copy-on-write `DELETE` means
"inherit the creation-time value", and the empty-string sentinel becomes the *only* way to say "none".

So I've accepted that loss for now: `DELETE` stays the clear operation and means "inherit". If the
product wants "per-job opt-out of a project glossary", it should be its own change. Today nobody can
express it either way — the settings weren't editable at all before this PR — so it isn't a
regression against `develop`. Happy to be argued out of this one; it's the only item where I went
against the review.

That drops your **Change 2** entirely, which is worth something on its own: the engines are in worse
shape for `''` than the review suggests. `MMT:141` already has the `is_string(…) && !== ''` guard
you're asking for (plus a truthy check downstream at `MMTServiceApi:533`), but `DeepL:125-139` has
**no guard at all** — `formality`, `glossary_id` and `model_type` go straight into `$parameters`,
which is `json_encode`d wholesale at `AbstractEngine:321`. So DeepL already sends
`"formality": null` for unset keys, and making `''` a legal stored value would make it send
`"formality": ""`. I'd rather not open that door for a state nothing asks for yet.

### One refinement to the Remove list

I kept `$jobScopedKeys` at `ProjectMetadataService:98` rather than removing it, and repurposed it: it
selects the *guard* now, not a skip. Your Change 1 can't just replace `!empty($engineValue)` wholesale, because
`!empty()` is the *correct* guard for the project-wide params — for `enable_mt_analysis` and
`mmt_activate_context_analyzer`, `0`/`false` is the absent default and shouldn't be persisted. It's
only wrong for the MT keys, where `mmt_ignore_glossary_case = false` and a threshold of `0` are real
answers. So:

```php
foreach ($extraKeys as $extraKey) {
    $engineValue = $projectStructure->$extraKey;

    if (isset($jobScopedKeys[$extraKey])) {
        // 0 and false are answers the owner chose for these keys, so !empty() cannot be used.
        // Ported from the removed JobCreationService::buildMtSettingsMetadata().
        if (!is_scalar($engineValue) || $engineValue === '') {
            continue;
        }
        $options[$extraKey] = is_bool($engineValue) ? (string)(int)$engineValue : (string)$engineValue;
        continue;
    }

    if (!empty($engineValue)) {
        $options[$extraKey] = $engineValue;
    }
}
```

Small note on framing: this guard is a cost the change creates rather than a fix the branch was
losing. On the branch as you reviewed it, the 11 MT keys never reached `!empty()` — they went through
`buildMtSettingsMetadata`'s guard instead. Moving them back to the project write is what re-exposes
them to it. Doesn't change the conclusion, just where the line item belongs. There are now two tests
pinning it: a threshold of `0` and `mmt_ignore_glossary_case = false` must both persist, and an empty
string must produce no row.

The threshold needed no special handling once the unset loop at `:110-114` went: it arrives via
`$options = $projectStructure->metadata` at `:34` and survives the `$allowedKeys` guard, since it's a
`ProjectsMetadataMarshaller` case.

### Section 4 is not an open question — it's a bug, and my docblock is wrong

Answering it definitively: **yes.** There's no second PHP connection — one DSN, built in
`Database::getConnection()` from the single `AppConfig::$DB_SERVER` — but ProxySQL sits in front of
the one there is, and the codebase documents the routing rule in four places, including
`JobCreationService:431`, in the very file this PR writes the job rows in:

> Re-querying through ProxySQL risks hitting a read replica that hasn't replicated the INSERT yet

plus `FastAnalysis:337` ("a bare `ping()` is a plain read a replica can answer while the master is
down"), and `ProjectCompletionService:90` and `:208`. The rule is: **any read not wrapped in
`$db->transaction()` may be answered by a replica.** `resolveJobAnalysisMetadata()` is called from
the publish loop at `:1240`, outside any transaction, so its job reads are replica-routable *and*
Redis-cached for 10 minutes.

Worse, my docblock at `:961-964` justifies the `null` project id with "read from the master node
inside a transaction … a cached read must not be allowed to replace it" — and that's
self-contradictory. The read at `:516-527` goes through `setCacheTTL(3600)`, so `DaoCacheTrait` can
serve it from Redis without issuing any SQL. `$projectMtQualityValue` is potentially an hour stale;
it was never the master-fresh value the comment claims.

Fixed independently of the design change: `resolveJobAnalysisMetadata()` now takes the project id
and passes it to `resolve()`, so the resolver's own project leg works, with `$projectMtQualityValue`
kept as the last resort — last rather than second precisely because it can be the staler of the two.
Precedence is job row → project row → the value `main()` materialised → `false`, and the docblock now
says that instead of the reverse. Three tests pin it, including one for the case that used to fall
through to 85.

One more thing while in there: because `:555` applies `?? 85` *before* the value is threaded down,
`$projectMtQualityValue` is never null on the `main()` path, so the `?? false` leg at `:996` is dead
code from `main()` — only reachable via a synthetic `_insertFastAnalysis()` call. Leaving that as-is
for now; flagging it so nobody reads the three-way `??` as doing more than it does.

### Seven things the review didn't cover

1. **`required: ["key"]`** — as above, `[{}]` was a 500, not a 400. Done, on all 16 branches.
2. **`lara_glossaries` and `mmt_glossaries` have no `minItems`**, so `[]` is writable — and for Lara
   it's forwarded verbatim: `Lara:316` tests `!== null`, and `lara-sdk/src/Translator.php:116`
   forwards whatever `getGlossaries()` returns. That's a live bug today, unrelated to the design
   question. Added `minItems: 1`, plus a `minimum` bound on `mmt_glossaries` items, which accepted
   `0` and negatives. `Lara.php` guards the forward now too, since the legacy project scope was never
   validated and can still hold an `[]`.
3. **`mt_quality_value_in_editor`'s default disagrees with itself.** `85` in analysis
   (`FastAnalysis:555`, `:1024`, `CreateProjectController:374`, `ProjectTemplateDao:85`) vs **`86`**
   in the editor (`GetContributionController:183`, `GetContributionRequest:99`). Both are consumed as
   `100 - $value`, so an unset threshold gave penalty 15 in analysis and 14 in the editor. Six
   hard-coded literals are now one `JobsMetadataMarshaller::DEFAULT_MT_QUALITY_VALUE`, settling on
   85. Flagging that this one is user visible: the editor's default penalty moves from 14 to 15.
4. **On merge, only the first chunk's rows survive** (`JobSplitMergeService:644-648`), so an MT edit
   made on a later chunk only is silently lost. Another small point for your request: under
   copy-on-write that degrades to the creation-time project value instead of the engine default.
5. **`project_metadata.value` is `varchar(2048)`; `job_metadata.value` is `text`.** Copy-on-write
   keeps the tighter cap on the base value for large glossary lists. Status quo rather than a
   regression, but worth knowing it's there.
6. **The split-propagation gap is wider than the PR body says.** `dialect_strict`,
   `mandatory_issues`, `public_tm_penalty` and `tm_prioritization` are all written at creation and
   none was in `propagatedOnSplit()`. All four added, and a test now asserts the list covers *every*
   declared key, so the next one added can't repeat it.
7. **GitGuardian's two findings are false positives** — `'fapw_' . bin2hex(random_bytes(4))` and
   `'intpw_' . …`, generated test job passwords. The check reports `skipping`; it needs a dashboard
   dismissal, not a code change.

### What landed

Merged `develop` (580 commits, 7 conflicts) and pushed two commits on top. State now:

| | |
| --- | --- |
| 1.1 rebase | done — `mergeable: MERGEABLE` |
| 1.2 per-key eviction | done, via `MetadataDao::destroyCacheForCredential()` on top of your new `destroyCache()` door. Two real-SQL regression tests, both verified to fail without it |
| 1.3 missing DAO argument | done, and the parameter is now non-nullable as you suggested |
| 1.4 schema `required` | done, on all 16 branches — plus `required: ["key"]`, which you had not asked for but which turns `[{}]` from a 500 into a 400 |
| §2 copy-on-write | done. 46 lines out of `JobCreationService`, the guard ported into the project write, both comments rewritten |
| §2.3 empty-string none | **not done** — see above; happy to be argued out of it |
| §4 FastAnalysis | done, treated as a bug rather than a question |
| N1, N2, N3, N6 | in — schema `required: ["key"]`, the glossary bounds, the threshold constant, the split propagation |
| N4, N5, N7 | documented, not fixed — the merge-path chunk-2 loss, the `varchar(2048)` cap, and the GitGuardian dismissal (which needs the dashboard, not code) |

PHPStan clean at level 8. Full suite: 10455 tests, 8 failures, all environmental and all pre-existing
on `develop` in my container — 4 assert ActiveMQ is unreachable while `matecat-amq` is up, and 4
confirmation-token cases fail because my local test DB still has `users.confirmation_token` as
`varchar(50)` where the checked-in schema says `varchar(255)`, so your scoped token truncates on
insert. Worth knowing if you hit the same four.

The frontend integration comment is updated: the "never both" guarantee is gone (it is now "may be
both, job wins"), though the rule the FE implements — read `job` first, fall back to `project` — is
unchanged, so no FE code changes. And the release-order constraint relaxes substantially: a new
project now carries the values on the project scope too, so today's UI keeps rendering the panel
correctly. Only the owner's *edits* need the UI to read `job` first, which makes shipping this first
a feature gap rather than a visible regression.

The PR body is rewritten to match — the Changes table had rows asserting the opposite of what the
code now does.

Still outstanding: the Bruno `dev` end-to-end pass, which I couldn't run from here.

PR #4785, opened 2026-08-25
✨ feat(icu): subfilter ICU segments with the ICU-compliant handlers in analysis and contribution

Ostico:
## Summary

A source segment carrying valid complex ICU syntax must be subfiltered with the ICU-compliant
handlers only, otherwise the ICU arguments come back wrapped in PH tags. The decision is a
property of the single segment, not of the project, so it is taken per segment on the stored
Layer 0 source — never on anything the client declares — and the resulting handler list travels
with every request that leaves for MyMemory, because MyMemory decodes what it receives with the
handlers it is given and re-encodes its matches with them.

This wires that decision through both paths that reach the engines: the analysis payload built
at project creation, and the editor contribution lookup.

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Changes

| File | Change |
|------|--------|
| `lib/Model/ProjectCreation/ProjectManager.php` | Picks the handler set per segment when building the fast-analysis payload, and carries the verdict on the row as `icu_source` so the worker ships the list the text was built with |
| `lib/Controller/API/App/GetContributionController.php` | Reads the stored contexts before the filter exists, takes the ICU decision on that source, builds the filter with it, and sends the same reduced list to the TM. The verdict is scoped to the branch that takes it, so PHPStan rejects any future reorder that would build the filter first |
| `lib/Utils/Subfiltering/IcuCompliantHandlers.php` | Reduction delegated to the library's new `HandlersSorter::resolveClassNames()`; result normalised to a list, since a non-sequential array would encode as a JSON object on the wire |
| `composer.json` / `composer.lock` | `matecat/subfiltering` `^4.1.1`, which exposes `resolveClassNames()` and `getOrderedHandlerTagNames()` |
| `tests/unit/Core/Controllers/GetContributionControllerTest.php` | ICU detection, the four wire-list cases, both context-conversion directions, and the no-drift invariant: the list the TM is told is the set the filter actually resolved |
| `tests/unit/Core/Model/ProjectCreation/FastAnalysisIcuSubfilteringTest.php` | Analysis payload: ICU kept as ICU, bare `{placeholder}` still PH-wrapped, flag off and invalid patterns fall back to the project handlers |

## Testing

```
OK (9718 tests, 31585 assertions)
[OK] No errors
```

PHPStan is run whole-tree; this repository has no baseline, so every reported error is a failure.
Diff coverage on the changed `lib/` lines measured at 22/22 = 100% against the 80% gate.

## AI Disclosure

Claude Code (claude-opus-5)

## Notes

`GetContributionWorker` deliberately needs no change: it forwards the
`subfiltering_handlers` it is given into both the TM and MT configs, so shipping the reduced
list from the controller is the whole fix.

The empty reduction travels as `null`, not `[]`: on the receiving side `[]` means "load the
default handlers" while `null` means "load none", so the two are opposite instructions.

Depends on `matecat/subfiltering` v4.1.1, already released and on Packagist.

Follow-up work, not in this PR:

- `SetContributionWorker` / `SetContributionMTWorker` have not been examined for the same
  handler-set question.
- The DB fallback `_getSegmentsForFastVolumeAnalysis` still sends raw Layer 0 with no flag.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "matecat/cwh": "^3.1",
         "matecat/klein": "^3.2",
         "matecat/simple-s3": "^v2.0.0",
-        "matecat/subfiltering": "^4.1",
+        "matecat/subfiltering": "^4.1.1",
         "matecat/whole-text-finder": "^2.0",
         "matecat/xliff-parser": "~v3.0",
         "matecat/icu-intl": "^v1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "22b2b2e05824fd2109a5f88fdeca63c4",
+    "content-hash": "4e99adbcc833cdc5ce0a69c2c91e9315",
     "packages": [
         {
             "name": "aws/aws-crt-php",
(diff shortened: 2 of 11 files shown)


Ostico:
Stacked follow-up: #4787 removes the `ICUSourceSegmentChecker` trait in favour of `ICUSourceSegmentDetector`. It is based on this branch, so merge this one first.

riccio82 (approved):

